PR #1, opened 2016-01-23
make it so parsing features can be turned on and off with configuration

bcoe:
- [x] enable/disable boolean groups `-cats`.
- [x] enable/disable camel-case expansion.
- [x] enable/disable dot notation.
- [x] enable/disable number parsing.
- [x] enable/disable boolean negation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,8 @@
+var assign = require('lodash.assign')
 var camelCase = require('camelcase')
 var path = require('path')
+var parentModule = require('parent-module')
+var readPkgUp = require('read-pkg-up')
 var tokenizeArgString = require('./lib/tokenize-arg-string')
 var util = require('util')
 
@@ -10,6 +13,7 @@ function parse (args, opts) {
   args = tokenizeArgString(args)
   // aliases might have transitive relationships, normalize this.
   var aliases = combineAliases(opts.alias || {})
+  var configuration = loadConfiguration()
   var defaults = opts.default || {}
   var envPrefix = opts.envPrefix
   var newAliases = {}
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -538,10 +546,19 @@ function parse (args, opts) {
...
     aliases: flags.aliases,
-    newAliases: newAliases
+    newAliases: newAliases,
+    configuration: configuration
   }
 }
 
+function loadConfiguration () {
+  var pkg = readPkgUp.sync({cwd: path.dirname(parentModule)}).pkg
+  return assign({
```
sindresorhus: I was thinking about adding a `defaults` option to https://github.com/sindresorhus/pkg-conf. Would that make you able to use it?

bcoe:
@sindresorhus yeah, a `defaults` stanza would be great -- I could stop pulling in `lodash.assign`.

The other issue I'm still not sure I've figured out the solution to is, in `npm@2.x` how will I be sure that configuration is loaded from the parent project, rather than `node_modules/yargs/package.json`, since the directory structure will look like this:

`node_modules/yargs/node_modules/yargs-parser`

sindresorhus:
> yeah, a defaults stanza would be great -- I could stop pulling in lodash.assign.

Cool. Added in v1.1.0. https://github.com/sindresorhus/pkg-conf#pkgconfsyncnamespace-options

> The other issue I'm still not sure I've figured out the solution to is, in npm@2.x how will I be sure that configuration is loaded from the parent project, rather than node_modules/yargs/package.json, since the directory structure will look like this:

You need to pass down `__filename` from where you want the parent to be read from (see docs). In this case, in `yargs`. I could have implemented a `levels` option, but I feel explicitly passing down the `__filename` is less fragile.

bcoe:
@sindresorhus see my last commit, that approach seems to work well for what I was imagining.

It would be nice if the defaults were applied even if a package.json can't be found with the given config stanza.

sindresorhus:
> that approach seems to work well for what I was imagining.

Pretty sure that will fail in some cases, but up to you.

> It would be nice if the defaults were applied even if a package.json can't be found with the given config stanza.

Done in v1.1.1. You can remove the boilerplate now.

nexdrew:
Just a sideline commentary - it occurs to me that, as of bcoe/yargs#343, we may now be looking up a parent `package.json` file in `yargs` and in `yargs-parser`. Granted, it's for two different reasons and in two different modules, but they should actually resolve to the same content, so it's potential double duty.

Perhaps we should consider combining efforts. Since `yargs-parser` is meant to work as a standalone module, it possibly makes sense to do the lookup here and expose the resolved `package.json` contents in the public API for consumption by `yargs`. Either that, or allow `yargs` to pass the resolved configuration to `yargs-parser` explicitly. We should keep in mind that performing the lookup in `yargs-parser` possibly adds unnecessary coupling to `yargs` that could instead be represented by an explicit contract of a public API.

PR #2, opened 2016-01-29
switch to app-root-path module

bcoe:
the [app-root-path](https://www.npmjs.com/package/app-root-path) module is well tested, and seems to bring us closer to the behavior I was looking for.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 var camelCase = require('camelcase')
-var findUp = require('find-up')
+var rootPath = require('app-root-path')
 var pkgConf = require('pkg-conf')
 var path = require('path')
 var tokenizeArgString = require('./lib/tokenize-arg-string')
@@ -12,7 +12,7 @@ function parse (args, opts) {
   args = tokenizeArgString(args)
   // aliases might have transitive relationships, normalize this.
   var aliases = combineAliases(opts.alias || {})
-  var configuration = loadConfiguration(opts.cwd || path.resolve(path.dirname(__filename), '../'))
+  var configuration = loadConfiguration(opts.cwd || rootPath.path)
   var defaults = opts.default || {}
   var envPrefix = opts.envPrefix
   var newAliases = {}
@@ -563,7 +563,7 @@ function loadConfiguration (cwd) {
       'parse-numbers': true,
       'boolean-negation': true
     },
-    cwd: findUp.sync('package.json', {cwd: cwd})
+    cwd: cwd
   })
 }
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -12,7 +12,7 @@ function parse (args, opts) {
   args = tokenizeArgString(args)
   // aliases might have transitive relationships, normalize this.
   var aliases = combineAliases(opts.alias || {})
-  var configuration = loadConfiguration(opts.cwd || path.resolve(path.dirname(__filename), '../'))
+  var configuration = loadConfiguration(opts.cwd || rootPath.path)
```
jameswomack: (it looks like) you don't need to even use app-root-path because pkg-conf uses find-root and find-root uses some of the same logic app-root-path uses: https://github.com/jden/find-root/blob/master/index.js#L5

sindresorhus: @jameswomack No, it uses https://github.com/sindresorhus/find-up internally. Though I honestly don't see the point of the switch as AFAIK `app-root-path` does the same thing as `pkg-conf` and walks up to the closest package.json.

bcoe:
@sindresorhus @jameswomack I was reading through the documentation for app-root-path and it sounds like there are some edge-cases worth taking into account: the alternatively named -.node_modules folder, globally installed modules. It seemed like it might be worthwhile to use a module that handles any edge-cases we might run into ... @sindresorhus you said the original implementation might miss some edge-cases? I'd like to avoid this.

sindresorhus:
> the alternatively named -.node_modules folder, globally installed modules.

That only applies when using `NODE_PATH`, which is officially highly discouraged, and ideally no tool should support its usage.

> you said the original implementation might miss some edge-cases? I'd like to avoid this.

Yes, and I presented IMHO the correct solution, which you declined. This does the same as the original implementation.

bcoe:
@sindresorhus gotcha, I think the existing approach is close to what you suggested then; it defaults to __filename, but accepts a cwd. Yargs in turn passes in its own __filename as cwd. I'm happy to stick with the existing approach if you think __filename covers most edge-cases.

What edge-cases do you think yargs passing in its __filename will miss? Do you think yargs itself needs a new config option to set cwd? I'd rather avoid this,

bcoe:
Note that I resolve to __filename ../, because I want to pull in the package.json using yargs; not yargs.

index.js
jameswomack: My bad @sindresorhus, I ended up on pkg-config instead of pkg-conf on NPM

PR #5, opened 2016-02-02
Add new type 'number'

maxrimue:
This adds a new type `number`.

A `number`-type option has the following special behaviour:
- If the flag itself is not provided, its value will default to `undefined`, _unless_ defined otherwise via `.default()`
- If no number is provided as a value, its value will default to `NaN` (that means, if `Number()` isn't capable of coercing the value to a `number` primitive)

_Note_: Where should I implement the logic for checking if the given value is a number? I thought of the following code:
```javascript
if (typeof x === 'number') {
  return x
} else {
  return Number(x) // Will return NaN if x is not coercible to a number
}
```
I am aware of `isNumber()`, however, it only returns true if the option `parse-numbers` equals `false`, which is obviously not related to the `number`-type. What do folks think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,7 @@ function parse (args, opts) {
     arrays: {},
     bools: {},
     strings: {},
+    numbers: {},
     counts: {},
     normalize: {},
     configs: {},
@@ -43,6 +44,10 @@ function parse (args, opts) {
 
   ;[].concat(opts.string).filter(Boolean).forEach(function (key) {
     flags.strings[key] = true
+  })
+
+  ;[].concat(opts.number).filter(Boolean).forEach(function (key) {
+    flags.numbers[key] = true
   })
 
   ;[].concat(opts.count).filter(Boolean).forEach(function (key) {
@@ -522,6 +527,7 @@ function parse (args, opts) {
     var def = {
       boolean: true,
       string: '',
+      number: undefined,
       array: []
     }
 
@@ -533,6 +539,7 @@ function parse (args, opts) {
     var type = 'boolean'
 
     if (flags.strings && flags.strings[key]) type = 'string'
+    else if (flags.numbers && flags.numbers[key]) type = 'number'
     else if (flags.arrays && flags.arrays[key]) type = 'array'
 
     return type
--- a/index.js
+++ b/index.js
@@ -546,6 +546,7 @@ function parse (args, opts) {
   }
 
   function isNumber (x) {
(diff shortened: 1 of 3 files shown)


test/yargs-parser.js
```diff
@@ -176,6 +176,20 @@ describe('yargs-parser', function () {
     x.should.be.a('string').and.equal('56')
   })
 
+  it('should default numbers to undefined', function () {
+    var n = parser([ '-n' ], {
+      number: ['n']
+    }).n
+    expect(n).to.equal(undefined)
```
bcoe: @nexdrew what are your thoughts about defaulting a number to `undefined` ... but one could also argue for `NaN` or `0`.

bcoe:
@maxrimue this is looking good so far, need any help getting tests passing?

index.js
```diff
@@ -533,12 +539,14 @@ function parse (args, opts) {
...
 
     if (flags.strings && flags.strings[key]) type = 'string'
+    else if (flags.numbers && flags.numbers[key]) type = 'number'
     else if (flags.arrays && flags.arrays[key]) type = 'array'
 
     return type
   }
 
   function isNumber (x) {
+    if (flags.numbers && flags.numbers[x] && !isNaN(x)) return true
```
bcoe: I think you can get rid of this check, it won't actually do anything, instead update:

https://github.com/yargs/yargs-parser/blob/master/index.js#L315

To be something like:

```js
var value = val
if (!checkAllAliases(key, flags.strings)) {
  if (isNumber(val) || checkAllAliases(key, flags.numbers)) value = Number(val)
}
```

lrlna: @bcoe got this going so far: 
```js
    if (!checkAllAliases(key, flags.strings)) {
      if (isNumber(val) || checkAllAliases(key, flags.numbers)) value = Number(val)
      if (!isNumber(val) && checkAllAliases(key, flags.numbers)) value = NaN
    }
```

The issue is that `isNumber(undefined)` returns false. I am wondering if we ought to change that to `return true` instead, or try another work around?

test/yargs-parser.js
lrlna: defaulting to `undefined` is better imo. Dealing with `NaN` has always been a pain in the butt in my experience.

lrlna:
@maxrimue sorry -- I can't push to your branch for some reason!! I am opening a new Pull Request #7 to add that line of code.

PR #7, opened 2016-02-14
New type number

lrlna:
This PR builds on top of @maxrimue #5 ; I wasn't able to push to his fork, so creating a new one. 

Summary: checks if value is not a number and assigns a `NaN`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,6 +31,7 @@ function parse (args, opts) {
     arrays: {},
     bools: {},
     strings: {},
+    numbers: {},
     counts: {},
     normalize: {},
     configs: {},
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -538,18 +548,24 @@ function parse (args, opts) {
...
 
     if (flags.strings && flags.strings[key]) type = 'string'
+    else if (flags.numbers && flags.numbers[key]) type = 'number'
     else if (flags.arrays && flags.arrays[key]) type = 'array'
 
     return type
   }
 
   function isNumber (x) {
+    if (flags.numbers && flags.numbers[x] && !isNaN(x)) return true
```
bcoe: Can we nuke this line? I'm pretty sure it's not having an effect.

lrlna: done.

PR #8, opened 2016-02-17
Fix behaviour for arrays when passed with no value and as the first option

maxrimue:
This fixes [yargs/#383](https://github.com/bcoe/yargs/issues/383#issuecomment-185270189).

If passed as the first option, without grouping and value, an array option would not get defined as `eatArray()` does simply abort if it doesn't see any value passed. This fix makes `eatArray()` assign the array option the default value if no value is provided.

There might be a more elegant solution, though...

### DIFF
--- a/index.js
+++ b/index.js
@@ -289,9 +289,10 @@ function parse (args, opts) {
   // following it... YUM!
   // e.g., --foo apple banana cat becomes ["apple", "banana", "cat"]
   function eatArray (i, key, args) {
+    var start = i + 1
     for (var ii = i + 1; ii < args.length; ii++) {
       if (/^-/.test(args[ii])) {
-        if (args[ii - 1].replace('-', '') === key) {
+        if (ii === start) {
           setArg(key, defaultForType('array'))
         }
         break
--- a/index.js
+++ b/index.js
@@ -290,7 +290,12 @@ function parse (args, opts) {
   // e.g., --foo apple banana cat becomes ["apple", "banana", "cat"]
   function eatArray (i, key, args) {
     for (var ii = i + 1; ii < args.length; ii++) {
-      if (/^-/.test(args[ii])) break
+      if (/^-/.test(args[ii])) {
+        if (args[ii - 1].replace('-', '') === key) {
+          setArg(key, defaultForType('array'))
+        }
+        break
+      }
       i = ii
       setArg(key, args[ii])
     }
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -1110,6 +1110,13 @@ describe('yargs-parser', function () {
       parse.should.have.property('_').with.length(0)
     })
 
+    it('should default an array to an empty array if passed as first option followed by another', function () {
+      var result = parser(['-a', '-b'], {
+        array: 'a'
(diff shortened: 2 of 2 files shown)


bcoe:
@maxrimue this looks reasonable, what was the behavior before?

maxrimue:
@bcoe The behaviour before this patch is detailed in the linked yargs issue. Basically, if you pass an array as a short option without any other added to it, pass no value, and pass at least one additional option, `eatArray()` is called and does not check for if any value is passed, which means that this:
```
myApp -a --someOption // Where 'a' is an array option
```
gets us `argv.a === undefined` instead of the desired `argv.a === []`

index.js
```diff
@@ -290,7 +290,12 @@ function parse (args, opts) {
   // e.g., --foo apple banana cat becomes ["apple", "banana", "cat"]
   function eatArray (i, key, args) {
     for (var ii = i + 1; ii < args.length; ii++) {
-      if (/^-/.test(args[ii])) break
+      if (/^-/.test(args[ii])) {
+        if (args[ii - 1].replace('-', '') === key) {
```
bcoe: sorry for the slow feedback, finally getting around to merging some things on yargs. I think it might be less fragile to simply use this check:

```js
if (ii === i + 1) {
}
```
ensuring that it's the first iteration of the loop.

maxrimue: @bcoe Won't this also return true in any other iteration? Since `i` gets assigned the value of `ii` at the end of each iteration, and `ii` gets incremented at the start of every iteration, doesn't that mean that `ii === i + 1` will always return true?

I still have passing tests, but I think that's because `yargs-parser` seems to only run arrays through `eatArray()` if they're short grouped and alone as the first option, like this:
```javascript
const yargs = require('yargs');

const argv = yargs
  .array('a')
  .array('b')
  .argv;
```
```
myModule -a -b
```
^ `eatArray()` is only run against `a`. If you call via `myModule -b -a` instead, it's run only against `b`. Is this intentional?

bcoe: whoops, you're right @maxrimue; I think we should add an extra test for this edge-case that I would have broke.

Could we instead do this:

```js
  // if an option is an array, eat all non-hyphenated arguments
  // following it... YUM!
  // e.g., --foo apple banana cat becomes ["apple", "banana", "cat"]
  function eatArray (i, key, args) {
    var start  = i + 1
    for (var ii = i + 1; ii < args.length; ii++) {
      if (/^-/.test(args[ii])) {
        if (ii === start) setArg(key, defaultForType('array'))
        break
      }
      i = ii
      setArg(key, args[ii])
    }

    return i
  }
```

mainly I think it reads clearer, you want to return `[]` if `0` positional arguments follow the starting key?

maxrimue: @bcoe Exactly, since that is what people expect as it's the default value for this type. I like the `start` idea, will update the PR

Alphapage:
Hello,

When do you plan to publish this fix to npm ?
I'm really waiting for this to solve my problems.

Thank you in advance.

maxrimue:
@Alphapage Sorry it took a bit, here we go! We will release a patch to npm soon

PR #9, opened 2016-02-21
Copy args before parsing

novemberborn:
When a flag is encountered with nargs set, and = is used, the input args array is modified. This breaks recursive command parsing in yargs itself.

Copy the args before parsing and add a test which shows the problem.

Fixes https://github.com/yargs/yargs/issues/396

### DIFF
--- a/index.js
+++ b/index.js
@@ -626,7 +626,7 @@ function increment (orig) {
 }
 
 function Parser (args, opts) {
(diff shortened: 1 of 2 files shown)


bcoe:
@novemberborn :+1: looks great, I'll work on getting a release out soon.

PR #10, opened 2016-03-16
Join 'keys' in hasKey() when 'dot-notation' is turned off

elas7:
I added a check for `configuration['dot-notation']` in `hasKey`, just like `setKey` already has. This way, if dot-notation is turned off, the keys will be joined. Therefore `hasKey` will be able to correctly tell if the key is in `obj`, without trying to check nested objects.

Without this fix, some errors occur when calling `applyDefaultsAndAliases` (which is the only method that calls `hasKey`) while dot-notation is disabled. For example:

- Values from the cli would fail to override defaults, if the defaults have dots.
- Values from the cli also fail to override aliases, if the aliases have dots.

### DIFF
--- a/index.js
+++ b/index.js
@@ -450,6 +450,9 @@ function parse (args, opts) {
 
   function hasKey (obj, keys) {
     var o = obj
+
+    if (!configuration['dot-notation']) keys = [keys.join('.')]
+
     keys.slice(0, -1).forEach(function (key) {
       o = (o[key] || {})
     })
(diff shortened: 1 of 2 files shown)


bcoe:
@elas7 lovely, thank you :) will work on getting this merged ASAP.

bcoe:
@elas7 what is your email address I would love to add you to our slack;

I'm ben@npmjs.com, if you would like to email it to me.

PR #13, opened 2016-03-28
Fix args.splice() when handling the case '--foo=xxx' with nargs or arrays

elas7:
When handling the `--f=a b c` case, if the key of `--f=a` is in 'nargs' or 'arrays', then the value 'a' is added as the next item in 'args'. 

There's an issue with the call used for that: `args.splice(i + 1, m[1], m[2])`, where `m[1]` is the key and `m[2]` is the value. There should be a zero in place of `m[1]`, because the signature of splice is `array.splice(start, deleteCount[, item1[, item2[, ...]]])`, and we don't actually try to delete anything from 'args'.

An edge case error occurs if the key is actually a number. If so, that number of items will be deleted from args. (For example: Calling with `--1=a b` will result in `1 = ["a", null]`).

Note than the args.splice call is already used correctly when handling the `-f=a b c` case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -121,11 +121,11 @@ function parse (args, opts) {
 
       // nargs format = '--f=monkey washing cat'
       if (checkAllAliases(m[1], flags.nargs)) {
-        args.splice(i + 1, m[1], m[2])
+        args.splice(i + 1, 0, m[2])
         i = eatNargs(i, m[1], args)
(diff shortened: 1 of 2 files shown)


bcoe:
@elas7 mind adding a failing test for this one edge-case? great fix.

elas7:
@bcoe Sure! Added.

PR #14, opened 2016-03-28
Improve code consistency when handling '-abc' case.

elas7:
I made some readability improvements:
- Replaced `letters[j]` with var `key`, in consistency with the rest of the code in that part.
- Replaced `arg.slice(j + 2)` with var `next`, in consistency with the rest of the code in that part.
- Created and used variable `next = args[i + 1]` when handling the last letter in 'arg'. This way it is more readable and looks exactly like the '--arg' case in line 150

### DIFF
--- a/index.js
+++ b/index.js
@@ -192,7 +192,7 @@ function parse (args, opts) {
           key = letters[j]
 
           // nargs format = '-f=monkey washing cat'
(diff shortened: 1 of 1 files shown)


bcoe:
:+1: great.

PR #15, opened 2016-03-29
Allow nested options in config files

elas7:
It solves yargs/yargs#292.

`setConfig()` now checks the config object recursively, and calls `setArg()` for every nested key if that key isn't in argv or has the default value.


bcoe:
awesome \o/

PR #17, opened 2016-04-02
Fix Boolean flag when next value contains the strings 'true' or 'false'

elas7:
When calling `test.js -b aaafalseaaa`, and `b` is a boolean, the value of `'b'` is set to `false` and `'aaafalseaaa'` doesn't appear in `'_'`.

I fixed this by updating the regex to look like the one in the `--foo true` case, where it works fine.


bcoe:
@elas7 mind rebasing 👍

bcoe:
great fix 👍

PR #19, opened 2016-04-03
Fix issue with numeric character in group of options

elas7:
When checking a group of options, if there is any number in the characters following the key being currently evaluated, then everything after the current key will be set as its value. This happens because the regex that checks for numbers is missing a `^` at the beginning, so strings like `'abc123'` are considered numbers.

An issue caused by this is that, for example, calling `-abc=123` results in `{"_": [], "a": "bc=123"}` instead of `{"_": [], "a": true, "b": true, "c": 123}`

### DIFF
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -762,6 +762,29 @@ describe('yargs-parser', function () {
       argv.should.have.property('n', 123)
     })
 
+    it('should set n to the numeric value 123, with n at the end of a group', function () {
+      var argv = parser([ '-ab5n123' ])
+      argv.should.have.property('a', true)
+      argv.should.have.property('b', true)
+      argv.should.have.property('5', true)
+      argv.should.have.property('n', 123)
+      argv.should.have.property('_').with.length(0)
+    })
+
+    it('should set n to the numeric value 123, with = as separator', function () {
+      var argv = parser([ '-n=123' ])
+      argv.should.have.property('n', 123)
+    })
+
+    it('should set n to the numeric value 123, with n at the end of a group and = as separator', function () {
+      var argv = parser([ '-ab5n=123' ])
+      argv.should.have.property('a', true)
+      argv.should.have.property('b', true)
+      argv.should.have.property('5', true)
+      argv.should.have.property('n', 123)
+      argv.should.have.property('_').with.length(0)
+    })
+
     it('should set option "1" to true, option "2" to true, and option "3" to numeric value 456', function () {
       var argv = parser([ '-123', '456' ])
       argv.should.have.property('1', true)
--- a/index.js
+++ b/index.js
@@ -212,8 +212,9 @@ function parse (args, opts) {
           continue
         }
 
+        // current letter is an alphabetic character and next value is a number
         if (/[A-Za-z]/.test(letters[j]) &&
-          /-?\d+(\.\d*)?(e-?\d+)?$/.test(next)) {
+          /^-?\d+(\.\d*)?(e-?\d+)?$/.test(next)) {
           setArg(letters[j], next)
           broken = true
           break


test/yargs-parser.js
```diff
@@ -762,6 +762,29 @@ describe('yargs-parser', function () {
       argv.should.have.property('n', 123)
     })
 
+    it('should set n to the numeric value 123, with n at the end of a group', function () {
+      var argv = parser([ '-ab5n123' ])
```
bcoe: part of me would expect this to equal:

a: true
b: true
5: true
n: true
1: true
2: true
3: true

test/yargs-parser.js
```diff
@@ -762,6 +762,29 @@ describe('yargs-parser', function () {
...
+    it('should set n to the numeric value 123, with = as separator', function () {
+      var argv = parser([ '-n=123' ])
+      argv.should.have.property('n', 123)
+    })
+
+    it('should set n to the numeric value 123, with n at the end of a group and = as separator', function () {
+      var argv = parser([ '-ab5n=123' ])
+      argv.should.have.property('a', true)
+      argv.should.have.property('b', true)
+      argv.should.have.property('5', true)
```
bcoe: I think, given the test above, I would expect:

`ab5n = 123`

bcoe:
@nexdrew mind lending @elas7 and I a second set of eyes?

test/yargs-parser.js
nexdrew: This is an odd edge case. I would expect something like `-x1` to be parsed as `x: 1`, which I think is common among most POSIX bins, but I don't know what the typical interpretation of `-x1y5` is.

In this test case, I would expect the result to be one of the following:

1. `a: true`, `b: 5`, `n: 123`
2. `a: true`, `b: true`, `5: true`, `n: 123`

test/yargs-parser.js
nexdrew: I kind of agree. This seems to be an alternate form of `--ab5n=123` since you can typically do `-abc=123` as an alternate to `--abc=123` (I think).

test/yargs-parser.js
bcoe: To me, option `.2.` feels a little more useful, I can't think of a situation where I want `5: true` but I can imagine a lazy person writing `-x1y5` and expecting:

`{x: 1, y:5}`

maxrimue: I would honestly expect it to be parsed like @bcoe noted in the first comment. I've actually not seen it in other bins, but I would expect grouping short options to not offer assigning values at all, because it's not really obvious to which options the values shall apply. As an example:
`-xy1` could mean:
1) Assign `1` to `x` and `y`, or
2) Assign `1` to `y` and `true` to `x`

Even if we would document it, it's just not obvious if you read something like this.
Additionally, currently we do assigning values to options with either a `=` or a space:
`--value=1` and `--value 1`
if we now allowed typing values _of a certain kind_  right after the name of the option, wouldn't that kinda break the consistency?

test/yargs-parser.js
elas7: @bcoe @nexdrew: Currently, you should get `ab5n = 123` only if the `'short-option-groups'` configuration is turned off. That way, the `-abc=123` and `--abc=123` are considered the same.

With `'short-option-groups'` turned on, I think it's expected behavior that every letter in a short option group is considered a separate option.

test/yargs-parser.js
elas7: I agree with @maxrimue that we should better not assign values on short option groups based on kinds of values, most of the time it is not really obvious what the user is trying to do.

Yet, the current behavior is that numbers at the end of an options group are set as the value of the last key. Maybe we could add a configuration to yargs-parser to toggle that off?

bcoe:
@elas7 you've swayed me in your argument for how parsing should be applied. I think it's worth kicking this change to the next major release of yargs however, it has the potential to change parsing behavior in a way that is hard for folks to debug.

PR #23, opened 2016-04-08
feat(configuration): Allow to pass configuration objects to yargs-parser

elas7:
This was brought to by https://github.com/yargs/yargs/issues/435, about aliases not being set correctly to values set by `pkgConf`. (It also solves https://github.com/yargs/yargs/issues/423).

For aliases to be handled correctly, even considering transitive relationships, `pkgConf` values should be handled in yargs-parser, where most of the logic of aliases is. So, the solution here is to pass the object from the key in `package.json` to yargs-parser.

This is the first part of a solution that would later need yargs to pass the object to yargs-parser.

Here are some design considerations I made:
- I thought the best is to add a generic key to `opts` in yargs-parser, called `configObjects` that takes arbitrary objects and adds them to `argv`. So as to avoid coupling it with `pkgConf`. That way it makes sense to people who use yargs-parser by itself.
- Also, I thought it's fine for the `configObjects` values to have almost the same priority that values set by a config file, after all, they are another way of configuration. An issue with that is that values from `pkgConf` were previously added as 'defaults', so they could have been overridden by env values before, but not anymore. Could this be an breaking change?

Let me know what you think. This is a WIP, if you are ok with the changes I'll add more tests and documentation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,6 +19,7 @@ function parse (args, opts) {
     'boolean-negation': true
   }, opts.configuration)
   var defaults = opts.default || {}
+  var configObjects = opts.configObjects || []
   var envPrefix = opts.envPrefix
   var newAliases = {}
   // allow a i18n handler to be passed in, default to a fake one (util.format).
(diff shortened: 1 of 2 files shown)


bcoe:
@elas7 awesome work, this will ultimately solve an issue I was running into with `nyc`.

index.js
```diff
@@ -436,6 +439,14 @@ function parse (args, opts) {
     })
   }
 
+  // set all config objects passed in opts
+  function setConfigObjects () {
+    if (typeof configObjects === 'undefined') return
```
nexdrew: The check for `undefined` seems to be handled by line 22: `opts.configObjects || []`.

Perhaps it would be better to make sure `configObjects` is an Array here:

```js
if (!Array.isArray(configObjects)) configObjects = [].concat(configObjects)
```

PR #26, opened 2016-04-10
feat(environment): Support nested options in environment variables

elas7:
It uses the '__' syntax to indicate nested properties, based on prior art from
https://github.com/indexzero/nconf#env and https://github.com/dominictarr/rc.

Closes yargs/yargs#332


bcoe:
@elas7 really dig this feature.

PR #29, opened 2016-04-21
Do not parse numbers that evaluate to Infinity

jviotti:
Some numbers, like `87e3202`, are recognised as numbers by the
isNumber utility, which results on `yargs-parser` parsing them by default
and thus passing a meaningless `Infinity` value to the user.

Since parsing scientific notation numbers are supported on `yargs-parser` as a
feature, the approach of this patch is to only return true from
`isNumber` if the string indeed represents a number, but it doesn't
evaluate to `Infinity` when being parsed as such, therefore not breaking
any existing functionality on the module.

The use case where this bug showed up was passing (shorter) UUIDs as
command line arguments.

Signed-off-by: Juan Cruz Viotti <jviottidc@gmail.com>

### DIFF
--- a/index.js
+++ b/index.js
@@ -602,7 +602,7 @@ function parse (args, opts) {
     if (!configuration['parse-numbers']) return false
     if (typeof x === 'number') return true
     if (/^0x[0-9a-f]+$/i.test(x)) return true
-    return /^[-+]?(?:\d+(?:\.\d*)?|\.\d+)(e[-+]?\d+)?$/.test(x)
+    return /^[-+]?(?:\d+(?:\.\d*)?|\.\d+)(e[-+]?\d+)?$/.test(x) && Number(x) !== Infinity
   }
 
   function isUndefined (num) {
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -185,6 +185,11 @@ describe('yargs-parser', function () {
     args.should.have.property('s', 'X\nX')
   })
 
+  it('should preserve numbers that evaluate to infinity as strings', function () {
+    var args = parser(['-n', '87e3202'])
+    args.should.have.property('n', '87e3202').and.be.a('string')
+  })
+
   it('should not convert numbers to type number if explicitly defined as strings', function () {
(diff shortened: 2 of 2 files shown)


nexdrew:
@jviotti Thanks for the suggestion!

Typically, the way to handle this use-case would be to:

1. Explicitly tell yargs-parser to parse the option as a string instead of a number via `opts.string`, e.g.

    ```js
    var parser = require('yargs-parser')
    var opts = { string: ['n'] }
    parser('-n 87e3202', opts)
    ```

2. Disable number coercion of strings that look like numbers via `opts.configuration['parse-numbers']`, e.g.

    ```js
    var parser = require('yargs-parser')
    var opts = {
      configuration: {
        'parse-numbers': false
      }
    }
    parser('-n 87e3202', opts)
    ```

And actually, `87e3202` is ambiguous since it could be a hex number (decimal `142488066`) or a big number in scientific notation or a string. And how should we know which one is meant unless we're told? Maybe a user really does mean the scientific notation, in which case `Infinity` is the correct interpretation.

Therefore, I'm not too sure this change is a good idea. If we do adopt this, then it's a breaking change to be sure (b/c explicitly parsing `87e3202` as a number via `opts.number` changes from `Infinity` to `NaN`).

Thoughts (from you and others)?

bcoe:
@jviotti I really appreciate the pull request, and debugging. But I think I agree with @nexdrew on this one. It's probably better to simply give the parser a hint, indicating that `-n` is type `string`.

Where this falls over slightly, is if you would like the values in `_` to be strings and they're being coerced into numbers. @nexdrew do we have a way to set a default value for `_`, seems like something we planned on doing.

nexdrew:
@bcoe @jviotti You can also tell yargs-parser to treat flagless/positional args as strings too via:

```js
var parser = require('yargs-parser')
var opts = { string: ['_'] }
parser('87e3202', opts)
```

This results in `_` being `[ '87e3202' ]` as expected.

jviotti:
Your suggestion looks reasonable! Thanks a lot!

PR #42, opened 2016-08-08
feat: added coerce option, for providing specialized argument parsing

bcoe:


nexdrew:
Since a callback is used, I would expect `yargs-parser` to support asynchronous coercion, but it doesn't (`yargs-parser` is fully synchronous). Values given to the callback asynchronously are ignored.

Perhaps `coerce` functions should either return a value or throw an error to make it clear that coercion must be synchronous.

nexdrew:
See #47

PR #46, opened 2016-08-08
allow non string values to be passed for values

lee-elenbaas:
allows for parsing of arrays like: ['--param', 15]

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,6 +115,7 @@ function parse (args, opts) {
     var letters
     var m
     var next
+    var nexts
     var value
 
     // -- seperated by =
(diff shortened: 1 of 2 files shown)


nexdrew:
Before we review this, would you mind describing the use-case(s) this will help solve apart from doing something like the following?

```js
var parse = require('yargs-parser')
parse(['-f', 11, '--zoom', 55].join(' ')) // <-- array to single string
parse(['-f', 11, '--zoom', 55].map(String)) // <-- array of strings
```

bcoe:
@lee-elenbaas I think I'd rather keep the explicit contract that `yargs-parser` expects an array of strings, does @nexdrew's suggestion of:

`parse(['-f', 11, '--zoom', 55].map(String))`

provide a reasonable workaround for you?

lee-elenbaas:
It is the workaround i am currently using - at the very least - that contract should be documented

nexdrew:
> at the very least - that contract should be documented

@lee-elenbaas Sounds reasonable to me. Thanks for bringing this up and for your understanding in this matter!

PR #47, opened 2016-08-08
refactor: expect coerce functions to be synchronous

nexdrew:
Expect a coercion function to return or throw instead of calling a callback, to avoid any ambiguity that async functions might be supported.


bcoe:
@nexdrew I like this API much better.

PR #58, opened 2016-09-28
feature: remove deprecated dependency lodas.assign

coderaiser:
When I install `yargs-parser` I see this warning:

`npm WARN deprecated lodash.assign@4.2.0: This package is deprecated. Use Object.assign.`

So I replaced `lodash.assign` with [object-assign](https://github.com/sindresorhus/object-assign).

`lodash` documentation [says](https://lodash.com/docs/4.16.2#assign):

```
This method mutates object and is loosely based on Object.assign.
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,4 @@
-var assign = require('lodash.assign')
+var assign = require('object-assign')
 var camelCase = require('camelcase')
 var path = require('path')
 var tokenizeArgString = require('./lib/tokenize-arg-string')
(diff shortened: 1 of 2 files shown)


bcoe:
@coderaiser thanks for bringing this to our attention, rather than pulling in a dependency I'm opting to simply implement this logic ourselves:

https://github.com/yargs/yargs-parser/pull/59

Since we're only using this logic to override default values, we can keep the logic pretty simple.

Hope that this is just the first of many contributions 👍

coderaiser:
@bcoe Great, that's even better :).

PR #60, opened 2016-10-01
Supportive changes for improved skipValidation functionality in `yargs`

faazshift:
See yargs/yargs#645 for more details.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,6 +21,7 @@ function parse (args, opts) {
   var configObjects = opts.configObjects || []
   var envPrefix = opts.envPrefix
   var newAliases = {}
+  var processedKeys = []
   // allow a i18n handler to be passed in, default to a fake one (util.format).
   var __ = opts.__ || function (str) {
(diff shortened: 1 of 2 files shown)


bcoe:
@nexdrew @faazshift my main concern with this pull and with https://github.com/yargs/yargs/pull/645, was that it introduces quite a bit of complexity to track which keys have been parsed by yargs vs. which keys came from config, environment, etc.

Mainly I see `setProcessedKey` being called in a lot of places, and I think when performing future maintenance of `yargs-parser`, we could quite easily break this functionality.

Do we still think it would be useful to be able to differentiate between which keys were processed by args, vs. were provided from environment, vs. config, etc? If this is a useful feature, I wonder if we could instead figure this out post-hoc using `process.argv`, `process.env`, and whatever config settings we have enabled?

CC: @nexdrew

bcoe:
haven't had feedback so I'm closing this for the time being; would love to open a conversation about it at some point though 👍

PR #63, opened 2016-10-15
fix: handle dot notation boolean options

ssonal:
Fixes yargs/yargs#617

On studying the code I found that this bug is being caused by yargs-parser. I think my fix should solve this issue, though I'm not sure if something else breaks because of these changes.

I've tested this modified library with yargs and found that all test cases pass.


bcoe:
@ssonal appreciate this patch, sorry for the slow turn around. Will hopefully have time to do a tiny bit of OSSing this weekend.

bcoe:
@ssonal thank you very much for this patch.

ssonal:
No problem!

PR #67, opened 2016-11-12
feat: allow disabling duplicate-arguments-array

laggingreflex:
For [yargs#530](https://github.com/yargs/yargs/issues/530)

Lets you disable "duplicates" behaviour

```
yargsParser('-x 1 -x 2')
// => {x: [1, 2]}
```
```
yargsParser('-x 1 -x 2', {configuration:{'duplicate-arguments-array': false}})
// => {x: 2}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,8 @@ function parse (args, opts) {
     'camel-case-expansion': true,
     'dot-notation': true,
     'parse-numbers': true,
-    'boolean-negation': true
+    'boolean-negation': true,
+    'duplicate-arguments-array': true
   }, opts.configuration)
   var defaults = opts.default || {}
   var configObjects = opts.configObjects || []
@@ -544,8 +545,10 @@ function parse (args, opts) {
       o[key] = value
     } else if (Array.isArray(o[key])) {
       o[key].push(value)
-    } else {
+    } else if (configuration['duplicate-arguments-array']) {
       o[key] = [ o[key], value ]
+    } else {
+      o[key] = value
     }
   }
 
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -1933,6 +1933,27 @@ describe('yargs-parser', function () {
         expect(parsed.dice).to.equal(undefined)
       })
     })
+
+    describe('duplicate arguments array', function () {
+      it('adds duplicate argument to array', function () {
+        var parsed = parser('-x a -x b', {
+          configuration: {
+            'duplicate-arguments-array': true
+          }
+        })
+
+        parsed['x'].should.deep.equal(['a', 'b'])
+      })
+      it('keeps only last argument', function () {
+        var parsed = parser('-x a -x b', {
+          configuration: {
+            'duplicate-arguments-array': false
+          }
+        })
+
+        parsed['x'].should.equal('b')
+      })
+    })
   })
 
   // addresses: https://github.com/yargs/yargs-parser/issues/41


index.js
```diff
@@ -544,8 +545,10 @@ function parse (args, opts) {
       o[key] = value
     } else if (Array.isArray(o[key])) {
       o[key].push(value)
-    } else {
+    } else if (configuration['duplicate-arguments-array']) {
```
bcoe: Great, that's about as simple as I was hoping it would be 👍

laggingreflex:
This has a slight problem though... it requires you to put `{"yargs": {"duplicate-arguments-array": false}}` in `package.json` [due to how yargs passes "configuration" to yargs-parser](https://github.com/yargs/yargs/blob/3fe9b8fdbd53da8841cc001cffe7f7794100031a/yargs.js#L824-L825). Any suggestions to specify this config elsewhere? Or should it be handled in yargs?

bcoe:
@laggingreflex you can also manually pass configuration to yargs-parser, which is how yargs actually does it 😄 

https://github.com/yargs/yargs-parser/blob/master/test/yargs-parser.js#L1777

☝️ you should be able to crib that test.

laggingreflex:
I might be missing something. Can the consumers of *yargs* be able to pass that configuration to yargs-parser though? Because I originally intended to fix [yargs#530](https://github.com/yargs/yargs/issues/530) with this.

```js
import yargs from 'yargs'

const config = yargs(???).parse('-x 1 -x 2');
```

It seemed to me the only way for a yargs user to pass this configuration would be by using package.json way - [due to how yargs passes "configuration" to yargs-parser](https://github.com/yargs/yargs/blob/3fe9b8fdbd53da8841cc001cffe7f7794100031a/yargs.js#L824-L825)

bcoe:
@laggingreflex yargs' is smart enough to read configuration from your projects own package.json; so once you add this setting, a user who wants this behavior, simply needs to add:

```{
  "yargs": {
     "duplicate-arguments-array": false
   }
}```

To their project's own package.json.

bcoe:
@laggingreflex I'm really happy with this, will work on getting this landed ASAP.

bcoe:
@laggingreflex mind rebasing, I think I landed your other pull request in the wrong order.

laggingreflex:
That one (#71) included this one (#67) so now both are applied, however the resultant commit (https://github.com/yargs/yargs-parser/commit/0f0fb2d13101ac1eb24479130a9290d5e431befc)'s message only reflects #71's changes. Would rebasing still help?

bcoe:
@laggingreflex if you checkout `master`, and rebase this branch against it, it'll do the write thing and make it so this can be landed -- If you don't, I can rebase on another branch for you, but it changes the commit history a bit.

laggingreflex:
This already has been landed.

#71 included commits for both this (#67) and  itself so as far as the matter is of landing this PR, it has already been done; it was included in #71 and the changes have landed successfully. 

The only problem is that the two commits have been squashed into one https://github.com/yargs/yargs-parser/commit/0f0fb2d13101ac1eb24479130a9290d5e431befc which makes no mention of this PR (#67), but only that of #71 (in the commit message). If it's not a big problem then you may close this PR. If however you do want separate commits then I think you probably will need to re-write history a bit. I did try to rebase but it gives me "nothing to commit", because like I said, the changed have already landed in latest master.

PR #75, opened 2016-12-02
fix: flatten/duplicate regression

### LINKED ISSUE #74: wrong behavior when flatten-duplicate-arrays and duplicate-arguments-array are enabled
@nexdrew pointed out this edge-case:

when you enable both  flatten-duplicate-arrays and duplicate-arguments-array are enabled

```json
{
  "yargs": {
    "flatten-duplicate-arrays": false,
    "duplicate-arguments-array": false
  }
}
```

Running:

```
node test.js -x 1 2 3 -x 2 3 4
```

results in:

```
{ _: [],
  x: [ [ '1', '2', '3' ], [ '2', '3', '4' ] ],
  '$0': 'test.js' }
```

It seems like it would be more consistent with this new API for the result to be `x: ['2', '3', '4']`

CC: @laggingreflex if you're looking for your next contribution to yargs, would love help adding a test and fix for this.

laggingreflex:
fixes #74 

Fix with tests for following possible cases:

        duplicate=false, flatten=false
          type=array
            [-x 1 2 3]          => [1, 2, 3]
            [-x 1 2 3 -x 2 3 4] => [2, 3, 4]
          type=string/number/etc
            [-x 1 -x 2 -x 3]    => 3

        duplicate=false, flatten=true
          type=array
            [-x 1 2 3]          => [1, 2, 3]
            [-x 1 2 3 -x 2 3 4] => [2, 3, 4]
          type=string/number/etc
            [-x 1 -x 2 -x 3]    => 3

        duplicate=true, flatten=true
          type=array
            [-x 1 2 3]          => [1, 2, 3]
            [-x 1 2 3 -x 2 3 4] => [1, 2, 3, 2, 3, 4]
          type=string/number/etc
            [-x 1 -x 2 -x 3]    => [1, 2, 3]

        duplicate=true, flatten=false
          type=array
            [-x 1 2 3]          => [1, 2, 3]
            [-x 1 2 3 -x 2 3 4] => [[1, 2, 3], [2, 3, 4]]
          type=string/number/etc
            [-x 1 -x 2 -x 3]    => [1, 2, 3]

The numbers are also preserved as numbers.

### DIFF
--- a/index.js
+++ b/index.js
@@ -325,8 +325,8 @@ function parse (args, opts) {
       i = ii
       argsToSet.push(args[ii])
     }
-    if (multipleArrayFlag && !configuration['flatten-duplicate-arrays']) {
-      setArg(key, argsToSet)
+    if (multipleArrayFlag) {
+      setArg(key, argsToSet.map(arg => processValue(key, arg)))
     } else {
       argsToSet.forEach(function (arg) {
         setArg(key, arg)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -325,8 +325,8 @@ function parse (args, opts) {
       i = ii
       argsToSet.push(args[ii])
     }
-    if (multipleArrayFlag && !configuration['flatten-duplicate-arrays']) {
-      setArg(key, argsToSet)
+    if (multipleArrayFlag) {
+      setArg(key, argsToSet.map(arg => processValue(key, arg)))
```
nexdrew: We want to maintain Node 0.10 support for now (until early next year), so the arrow function needs to be replaced with a plain old function.

index.js
```diff
@@ -551,14 +556,142 @@ function parse (args, opts) {
...
       o[key] = value
     } else if (Array.isArray(o[key])) {
-      o[key].push(value)
+      if (!duplicate && !flatten) {
+        if (isTypeArray) {
+          if (isOldValueArray) {
+            if (isNewValueArray) {
+              o[key] = value
+            } else {
+              o[key] = o[key].concat([value])
```
nexdrew: This logic can be simplified.

index.js
```diff
@@ -551,14 +556,142 @@ function parse (args, opts) {
...
+            }
+          } else {
+            if (isNewValueArray) {
+              throw new Error('this should not happen')
```
nexdrew: These are unnecessary and drop us below 100% test coverage.

bcoe:
@laggingreflex I really like these new configuration settings, they're a power feature though (for folks with needs like yourself) so I'd like to see the changes to support them be as minimal as possible.

@nexdrew's fixes are closer to the amount of code-changes I was thinking; better that we check the config flags a couple more times than significantly change the processing logic.

I'm excited about these new features was played with them in yargs yesterday.

laggingreflex:
Thanks @nexdrew! I was fully expecting a review for further simplification of all that unnecessary stuff. Let me know if it needs more work.

bcoe:
@nexdrew what's the status of this; would love to see this over the finish line.

bcoe:
@laggingreflex thank you! appreciate all the additional tests.

PR #78, opened 2017-02-08
Also flatten-duplicate-arrays when used with dot-notation

JeroenVdb:
Background: https://github.com/yargs/yargs/issues/777

When using dot-notations I also want arrays to be flattened. This was the behaviour in Yargs 5 but not anymore in Yargs 6.

```
# GOOD
$ node good-yargs-parser.js --metric.list AA --metric.list BB --metricList CC --metricList DD
{ _: [],
  metric: { list: [ 'AA', 'BB' ] },
  metricList: [ 'CC', 'DD' ],
  '$0': 'yargstest.js' }

# BAD
$ node bad-yargs-parser.js --metric.list AA --metric.list BB --metricList CC --metricList DD
{ _: [],
  metric: { list: [ 'AA', ['BB'] ] },
  metricList: [ 'CC', 'DD' ],
  '$0': 'yargstest.js' }
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -556,9 +556,16 @@ function parse (args, opts) {
       o = o[key]
     })
 
+    var fullKey = []
+    if (configuration['flatten-duplicate-arrays'] && keys.length > 1) {
+      fullKey.push(keys.join('.'))
+    } else {
+      fullKey.push(keys[keys.length - 1])
+    }
+
     var key = keys[keys.length - 1]
 
-    var isTypeArray = checkAllAliases(key, flags.arrays)
+    var isTypeArray = checkAllAliases(fullKey, flags.arrays)
     var isValueArray = Array.isArray(value)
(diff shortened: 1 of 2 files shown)


JeroenVdb:
@bcoe I don't think I understand correct. Doesn't my patch restore the previous 5.X behaviour?

bcoe:
@JeroenVdb I just didn't like that we were starting to check `foo.bar` for object variables ... perhaps this is the minimal change to get behavior back to a reasonable spot though 👍 

I'm going to be at [Hack Illinois](https://hackillinois.org/) this coming weekend, working with students on yargs, expect a lot of things to land (also, this is a good time to get additional issues open).

bcoe:
@JeroenVdb closing in favor of: https://github.com/yargs/yargs-parser/pull/86 (which I believe solves the more general case).

PR #81, opened 2017-02-18
fix: environment variables should take precedence over config file

bcoe:
BREAKING CHANGE: environment variables now take precedence over config file values.

see: https://github.com/yargs/yargs/issues/627

CC: @jaridmargolin


test/yargs-parser.js
```diff
@@ -1708,7 +1708,7 @@ describe('yargs-parser', function () {
         }
       })
 
-      result.herp.should.equal('derp')
```
kennethalegre19, line 1711: zerp

PR #83, opened 2017-03-08
Merge with config when array

markbirbeck:
This is a fix for yargs/yargs#817.

### DIFF
--- a/index.js
+++ b/index.js
@@ -468,7 +468,7 @@ function parse (args, opts) {
       } else {
         // setting arguments via CLI takes precedence over
         // values within the config file.
-        if (!hasKey(argv, fullKey.split('.')) || (flags.defaulted[fullKey])) {
+        if (!hasKey(argv, fullKey.split('.')) || (flags.defaulted[fullKey]) || (flags.arrays[fullKey])) {
           setArg(fullKey, value)
         }
       }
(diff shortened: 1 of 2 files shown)


markbirbeck:
Shouldn't the coverage test just check the entire branch, rather than individual steps? I'm assuming that the coverage number appears to fall because I add my test in one commit before I add the code that passes that test in the next. This makes it easier to review.

bcoe:
@markbirbeck thanks for the contribution (regarding coverage, this is a change in how coveralls works will not block the merging of this pull request).

markbirbeck:
Thanks for the update @bcoe.

bcoe:
@markbirbeck sorry to drop this on the floor for so long -- just working through a large backlog of issues now; having slept on this one. It comes to mind that we might want to make this a parser option that can be enabled or disabled. Here's a scenario I frequently run into:

* I'm configuring a cluster of servers, which are represented by an array; ['staging-www-1, 'staging-www-2] -- I might have this stored in a `config.json`.
* In production, I might wish to override this value with `['prod-www-1', 'prod-www-2']` -- with this new logic we'd instead end up with `['staging-www-1', 'staging-www-2', 'prod-www-1', 'prod-www-2']`?

Perhaps we simply add a new config option, `combine-arrays`, which defaults to `false`; if turned on it would combine merge arrays:

* if multiple arrays are given as arguments: `--arr 1 2 --arr 3 4`.
* or if a config file is provided along with an argument.

bcoe:
@markbirbeck does my last comment make sense? still interested in landing this?

PR #88, opened 2017-04-15
feat: when parsing stops, we now populate "--" by default

bcoe:
BREAKING CHANGE: rather than placing arguments in "_", when parsing is stopped via "--"; we now populate an array called "--" by default.

I've changed the configuration option to be named `populate--` rather than `--`; as this better fits with the naming conventions of our other configuration settings.

CC: @ruimarques

### DIFF
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -2228,9 +2228,7 @@ describe('yargs-parser', function () {
           '--key', 'value',
           '-b', '--bool', '--no-meep', '--multi=baz',
           '--', '--not-a-flag', '-', '-h', '-multi', '--', 'eek'
-        ], {
-          '--': true
-        })
+        ])
         result.should.have.property('c', true)
         result.should.have.property('a', true)
         result.should.have.property('t', true)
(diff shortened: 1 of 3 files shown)


bcoe:
@JaKXz whoops, I forgot to drop 0.12/0.12 from AppVeyor; I figured since it's yargs 8.x, we should drop Node 0.10/0.12.

test/yargs-parser.js
```diff
@@ -2233,6 +2197,44 @@ describe('yargs-parser', function () {
...
+
+      it('should populate the "--" array by default', function () {
+        var result = parser([
+          '--name=meowmers', 'bare', '-cats', 'woo', 'moxy',
+          '-h', 'awesome', '--multi=quux',
+          '--key', 'value',
+          '-b', '--bool', '--no-meep', '--multi=baz',
+          '--', '--not-a-flag', '-', '-h', '-multi', '--', 'eek'
+        ], {
+          '--': true
```
ruimarques: Shouldn't this be `configuration: {'populate--': true}` ?

bcoe: whoops, I should actually just remove this configuration; since it defaults to `true` now.

PR #96, opened 2017-08-16
Fix duplicate option values when combined with its equivalent dashed …

### LINKED ISSUE #95: Camel case option name returns an array when combined with its equivalent dashed option name
When running the following
```js
        var result = parser(['--someOption', 'val'], {
          alias: {
            someOption: 'some-option'
          }
        })
```
someone would expect 
```{ _: [], someOption: 'val', 'some-option': 'val' }```
as a result, but instead will receive 
```{ _: [], someOption: [ 'val', 'val' ], 'some-option': 'val' }```

The issue affects also [yargs](https://github.com/yargs/yargs/issues/920)

### PR DESCRIPTION
**Pull request by antoniom**
Fixes #95

### DIFF
--- a/index.js
+++ b/index.js
@@ -581,7 +581,7 @@ function parse (args, opts) {
       }
     } else if (o[key] === undefined && isTypeArray) {
       o[key] = isValueArray ? value : [value]
-    } else if (duplicate && !(o[key] === undefined || checkAllAliases(key, flags.bools) || checkAllAliases(keys.join('.'), flags.bools) || checkAllAliases(key, flags.counts))) {
+    } else if (duplicate && !(o[key] === undefined || o[key] === value || checkAllAliases(key, flags.bools) || checkAllAliases(keys.join('.'), flags.bools) || checkAllAliases(key, flags.counts))) {
       o[key] = [ o[key], value ]
     } else {
       o[key] = value
--- a/index.js
+++ b/index.js
@@ -581,7 +581,7 @@ function parse (args, opts) {
       }
     } else if (o[key] === undefined && isTypeArray) {
       o[key] = isValueArray ? value : [value]
-    } else if (duplicate && !(o[key] === undefined || o[key] === value || checkAllAliases(key, flags.bools) || checkAllAliases(keys.join('.'), flags.bools) || checkAllAliases(key, flags.counts))) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by bcoe** [comment]
@antoniom thanks for the patch 👍 will work on getting this landed soon.

Feel free to join: http://devtoolscommunity.herokuapp.com/

If you want to help move the release process along.

File: index.js [hunk 1]
```diff
@@ -581,7 +581,7 @@ function parse (args, opts) {
       }
     } else if (o[key] === undefined && isTypeArray) {
       o[key] = isValueArray ? value : [value]
-    } else if (duplicate && !(o[key] === undefined || checkAllAliases(key, flags.bools) || checkAllAliases(keys.join('.'), flags.bools) || checkAllAliases(key, flags.counts))) {
+    } else if (duplicate && !(o[key] === undefined || o[key] === value || checkAllAliases(key, flags.bools) || checkAllAliases(keys.join('.'), flags.bools) || checkAllAliases(key, flags.counts))) {
```
bcoe: this would introduce issues for folks who are trying to create an array like so:

```
parser.js -x 9 -x 9
```

I believe, I think a more general solution would be to use logic similar to this:

https://github.com/yargs/yargs-parser/blob/master/index.js#L352

and introduce [decamelize](https://www.npmjs.com/package/decamelize), so that we handle both going from `foo-bar` to `fooBar` and `fooBar` to `foo-bar`.

bcoe:
@antoniom interested in seeing this over the finish line, I think this would be great to fix. Perhaps you could join http://devtoolscommunity.herokuapp.com/ if you want any clarification about the comments I made 👍 thanks for the contributions so far.

antoniom:
Yes; I will try to provide a fix as soon as I have some time to check it!
I am really busy these days!

bcoe:
@antoniom no rush, appreciate your contributions when possible.

antoniom:
@bcoe I am not sure what should I refactor so that `decamelize` is used. 
For that reason I have prepared a patch so that camel cased keys are not added to `flags.aliases` when the came case version is identical to `key`. It seems OK, at least for the provided test cases plus the one you mentioned on your commend before (i.e. `parser.js -x 9 -x 9`)

bcoe:
@antoniom thank you for this patch; I apologize for taking so long to land, and not noticing the edits made (your follow up patch addressed my concerns surrounding array handling).

PR #97, opened 2017-08-17
Do not remove quotes when tokenizing a non-option string

### LINKED ISSUE #93: yargs not passing string correctly
```
describe('TokenizeArgString', function () {
  it.only('....handles quoted string', function () {
    var args = tokenizeArgString('-q sku="invalid"')
    args[0].should.equal('-q') // passes
    args[1].should.equal('sku="invalid"') //fails
  })
})
```

This test fails on this module. Not sure how to fix it.

antoniom:
Fixes #93

### DIFF
--- a/lib/tokenize-arg-string.js
+++ b/lib/tokenize-arg-string.js
@@ -2,7 +2,7 @@
 module.exports = function (argString) {
   if (Array.isArray(argString)) return argString
 
-  argString = argString.trim();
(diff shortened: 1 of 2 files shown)


bcoe:
@antoniom thank you very much for the patch 👋 

would love to have you join the community chat if you have any interest in chipping in more 😄 

PR #100, opened 2017-08-29
Fix: Ignore multiple spaces between arguments.

### LINKED ISSUE #99: Multiple spaces should be ignored unless they are quoted.
If multiple spaces are found between arguments, currently zero length arguments are inserted between each space. Instead they should be ignored.

See: trufflesuite/truffle#399

Mike111177:
Fixes #99

### DIFF
--- a/lib/tokenize-arg-string.js
+++ b/lib/tokenize-arg-string.js
@@ -3,16 +3,20 @@ module.exports = function (argString) {
   if (Array.isArray(argString)) return argString
 
   var i = 0
+  var prevC = null
   var c = null
   var opening = null
   var args = []
 
   for (var ii = 0; ii < argString.length; ii++) {
+    prevC = c
     c = argString.charAt(ii)
 
     // split on spaces unless we're in quotes.
     if (c === ' ' && !opening) {
-      i++
+      if (!(prevC === ' ')) {
+        i++
+      }
       continue
     }
 
--- a/test/tokenize-arg-string.js
+++ b/test/tokenize-arg-string.js
@@ -3,6 +3,7 @@
 var tokenizeArgString = require('../lib/tokenize-arg-string')
 
 require('chai').should()
+var expect = require('chai').expect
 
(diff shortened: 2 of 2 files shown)


bcoe:
@Mike111177 hey 👋 do to some other potential parsing issues I thought of with the parsing around maintaining quotes, I opted to rollback the logic surrounding `&& (prevC === ' ' || args[i].startsWith('-')`.

tldr; since most command lines already strip `"` and `'` I'm tempted to advice people to think of alternative ways to pass quoted strings to an application, vs., trying to cleverly maintain quotes strings.

Do you mind potentially rebasing your work with master, taking into account that this logic has been removed?

Mike111177:
@bcoe So by that you mean you want me to remove the commit from Antonis correct?

So I have to admit, I am pretty novice with git, and I am struggling to get the rebase right. I have managed to roll back the forked master branch to match, however for fix-double-spaces branch, where my work is, I seem to be unable to rebase it in a way that maintains my work while removing the commit from Antonis. Hopefully, I can figure this out soon.

Mike111177:
@bcoe Done.

bcoe:
@Mike111177 thanks for the contribution.

bcoe:
@Mike111177 this is now released in `yargs-parser@8.0.0` mind trying it out and making sure it works for you?

Mike111177:
@bcoe It now works as I was hoping, multiple space between arguments are now ignored and undefined are not inserted in the middle of the token array. 

I did notice however that when the are spaces at the beginning of the argstring, it will still add undefined's to the beginning. I did not test it for endings, however it could be solved by doing a  trim() on the argstring before tokenizing it.

bcoe:
@Mike111177 👍 I would happily accept a patch for the follow on `.trim()`.

I have all of yargs' car parts spread out over the garage right now working on [this](https://github.com/yargs/yargs/pull/967), if you have any feedback on this front.

PR #103, opened 2017-09-06
fix: only run coercion functions once, despite aliases. (#76)

### LINKED ISSUE #76: A coerce function runs again for each alias
The coerce function for an argument with an alias will run multiple times, the initial run and then once more per alias. 

I only noticed as I was using the file coercion example and reading a largish file. 

```
let argv = require('yargs')
  .alias('file', 'f2')
  .alias('file', 'f3')
  .coerce('file', function (arg) {
    console.log('reading file', arg)
    return require('fs').readFileSync(arg, 'utf8')
  })
  .argv
```
then
```
$ node testyargs.js  --file test
reading file test
reading file test
reading file test
```

from https://github.com/yargs/yargs/issues/729

Mike111177:
Fixes #76 
Fixes yargs/yargs#923

### DIFF
--- a/index.js
+++ b/index.js
@@ -513,13 +513,19 @@ function parse (args, opts) {
 
   function applyCoercions (argv) {
     var coerce
+    var applied = {}
     Object.keys(argv).forEach(function (key) {
-      coerce = checkAllAliases(key, flags.coercions)
-      if (typeof coerce === 'function') {
-        try {
-          argv[key] = coerce(argv[key])
-        } catch (err) {
-          error = err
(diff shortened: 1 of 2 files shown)


bcoe:
@Mike111177 wouldn't this mean that if `foo` and `bar` are aliases, only one of either `foo` or `bar` has the appropriate final value? or am I missing something.

Mike111177:
@bcoe, no, note lines 523 and 524. Once a coerce function runs, it applies the value to all needed aliases. Then on the next rounds it ignores the aliases that have already been assigned to.

PR #104, opened 2017-10-04
feat: reworking how numbers are parsed

bcoe:
Begins to address issues discussed in https://github.com/yargs/yargs/issues/96, moving us towards

* don't magically parse numbers larger than `Number.MAX_SAFE_INTEGER`.
* pull number parsing logic into `maybeCoerceNumber`.
* fix broken `parse-numbers = false` logic.

### DIFF
--- a/index.js
+++ b/index.js
@@ -400,7 +400,7 @@ function parse (args, opts) {
       newAliases[alias] = true
     }
     if (!(flags.aliases[alias] && flags.aliases[alias].length)) {
-      addNewAlias(alias)
+      addNewAlias(alias, key)
     }
   }
 
--- a/index.js
+++ b/index.js
@@ -425,7 +425,7 @@ function parse (args, opts) {
 
   function maybeCoerceNumber (key, value) {
     if (!checkAllAliases(key, flags.strings) && !checkAllAliases(key, flags.coercions)) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -421,6 +423,14 @@ function parse (args, opts) {
     return value
   }
 
+  function maybeCoerceNumber (key, value) {
+    if (!checkAllAliases(key, flags.strings) && !checkAllAliases(key, flags.coercions)) {
+      const shouldCoerceNumber = isNumber(value) && configuration['parse-numbers'] && (Number(value) <= Number.MAX_SAFE_INTEGER)
```
ezzatron: It might be better to use [`Number.isSafeInteger()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger) instead of manually comparing to [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER). If you don't, you probably need to think about large *negative* numbers and using [`Number.MIN_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER) too.

Issue #107, opened 2017-11-14
setConfigObject() crashes on null value for object key 


zyrolasting:
PR on #108. Given the following entry point...

    require('yargs')
        .config({key: null})
        .commandDir(`${__dirname}/tasks`)
        .demandCommand()
        .help()
        .argv;

You get this stack trace:

    /home/sage/Code/redacted/node_modules/yargs/node_modules/yargs-parser/index.js:476
        Object.keys(config).forEach(function (key) {
               ^

    TypeError: Cannot convert undefined or null to object
        at Function.keys (<anonymous>)
        at setConfigObject (/home/sage/Code/redacted/node_modules/yargs/node_modules/yargs-parser/index.js:476:12)
        at /home/sage/Code/redacted/node_modules/yargs/node_modules/yargs-parser/index.js:485:9
        at Array.forEach (<anonymous>)
        at setConfigObject (/home/sage/Code/redacted/node_modules/yargs/node_modules/yargs-parser/index.js:476:25)
        at /home/sage/Code/redacted/node_modules/yargs/node_modules/yargs-parser/index.js:500:7
        at Array.forEach (<anonymous>)
        at setConfigObjects (/home/sage/Code/redacted/node_modules/yargs/node_modules/yargs-parser/index.js:499:19)
        at parse (/home/sage/Code/redacted/node_modules/yargs/node_modules/yargs-parser/index.js:286:3)
        at Function.Parser.detailed (/home/sage/Code/redacted/node_modules/yargs/node_modules/yargs-parser/index.js:776:10)
        at Object.parseArgs [as _parseArgs] (/home/sage/Code/redacted/node_modules/yargs/yargs.js:967:27)
        at Object.get [as argv] (/home/sage/Code/redacted/node_modules/yargs/yargs.js:957:21)
        at Object.<anonymous> (/home/sage/Code/redacted/ui:18:5)
        at Module._compile (module.js:624:30)
        at Object.Module._extensions..js (module.js:635:10)
        at Module.load (module.js:545:32)

It seems this happens because `typeof null === 'object'` and thus triggers recursion in `setConfigObject()`, namely on the condition in index.js:482

	// if the value is an inner object and we have dot-notation
	// enabled, treat inner objects in config the same as
	// heavily nested dot notations (foo.bar.apple).
	if (typeof value === 'object' && !Array.isArray(value) && configuration['dot-notation']) {
		// if the value is an object but not an array, check nested object
		setConfigObject(value, fullKey)
[...]

PR #108, opened 2017-11-14
Fix: Allow null config values

zyrolasting:
See #107


bcoe:
@zyrolasting great little catch 😄 thanks for the contribution.

PR #109, opened 2017-11-28
.idea directory and package-lock.json are added to .gitignore as it's done in yargs

ezze:
This pull request adds `.idea` directory and `package-lock.json` file to `.gitignore` as it's [done in yargs](https://github.com/yargs/yargs/blob/master/.gitignore).


bcoe:
@ezze thank you 👍

PR #110, opened 2017-12-18
fix: scientific notation circumvented bounds check

bcoe:
due to a bug in our bounds checking logic, scientific notation could be entered greater than MAX_INTEGER.

fixes: https://github.com/yargs/yargs/issues/1017
see: https://github.com/yargs/yargs/pull/1019


ogonkov:
This fix first appear in v8.1.0 release?

PR #114, opened 2018-01-20
feat: narg arguments no longer consume flag arguments

bcoe:
BREAKING CHANGE: arguments of form --foo, -abc, will no longer be consumed by nargs

This is in prep for replacing the crufty legacy method `requiresArg` with nargs(1).

CC: @evocateur for review


bcoe:
@evocateur I don't remember if I consciously made this design decision when building nargs, or if it was an oversight -- this behavior feels better to me.

PR #119, opened 2018-03-31
fix: do not set boolean flags if not defined in `argv`

### LINKED ISSUE #116: Do not set boolean to `false` if flag is not in args
Currently when defining a `boolean` option, without specifying its default value, if the user doesn't set the corresponding arg the value is set to `false`.

Therefore its impossible to know if users set the flag to `false` or this they just omitted it.

This is problematic in situation in which you want to merge the options returned by `yargs-parser` with another object options (from a config file for example) with `Object.assign` or similar.
Typically you would merge options so the CLI args take precedence  over the config file. So users can specify their config and override it in development mode by passing an option to the cli.

In such situation, if the user do not pass the flag in the CLI, the merged options will always have `false` for this flag, as `yargs-parser` option have it set to false and it override what's in the config.

Ideally the value of a boolean flag should be `undefined` when the user doesn't pass the flag in the CLI.

Would be open to a PR that implement such behavior?

pvdlg:
Fix #116

`boolean` flags defined without a `default` value will now behave like other option type and won't be set in the parsed results when the user doesn't set the corresponding CLI arg.

Previous behavior:
```js
var parse = require('yargs-parser');

parse('--flag', {boolean: ['flag']});
// => { _: [], flag: true }

parse('--no-flag', {boolean: ['flag']});
// => { _: [], flag: false }

parse('', {boolean: ['flag']});
// => { _: [], flag: false }
```

New behavior:
```js
var parse = require('yargs-parser');

parse('--flag', {boolean: ['flag']});
// => { _: [], flag: true }

parse('--no-flag', {boolean: ['flag']});
// => { _: [], flag: false }

parse('', {boolean: ['flag']});
// => { _: [] } => flag not set similarly to other option type
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -105,8 +105,10 @@ function parse (args, opts) {
   var argv = { _: [] }
 
   Object.keys(flags.bools).forEach(function (key) {
-    setArg(key, !(key in defaults) ? false : defaults[key])
-    setDefaulted(key)
+    if (Object.prototype.hasOwnProperty.call(defaults, key)) {
+      setArg(key, defaults[key])
+      setDefaulted(key)
+    }
   })
 
   var notFlags = []
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -199,7 +199,7 @@ describe('yargs-parser', function () {
       boolean: ['x', 'y', 'z']
     })
     parse.should.have.property('x', true).and.be.a('boolean')
-    parse.should.have.property('y', false).and.be.a('boolean')
+    parse.should.not.have.property('y')
     parse.should.have.property('z', true).and.be.a('boolean')
     parse.should.have.property('_').and.deep.equal(['one', 'two', 'three'])
   })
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -105,8 +105,10 @@ function parse (args, opts) {
   var argv = { _: [] }
 
   Object.keys(flags.bools).forEach(function (key) {
-    setArg(key, !(key in defaults) ? false : defaults[key])
-    setDefaulted(key)
+    if (key in defaults) {
```
ljharb: This should use `Object.prototype.hasOwnProperty.call`, in case the "key" is "toString" or "hasOwnProperty"

pvdlg: I kept `key in defaults` because if was there before, but I agree we should use `Object.prototype.hasOwnProperty.call`. I made the change.

pvdlg:
Awesome! Thanks!

pvdlg:
@bcoe could you by any chance also check out yargs/yargs#1071 which is related? Thanks!

PR #122, opened 2018-05-11
feat: add directory key type

bcoe:
BREAKING CHANGE: fs is now a dependency

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,8 @@
-var camelCase = require('camelcase')
-var path = require('path')
-var tokenizeArgString = require('./lib/tokenize-arg-string')
-var util = require('util')
+const camelCase = require('camelcase')
+const path = require('path')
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -1,7 +1,8 @@
-var camelCase = require('camelcase')
-var path = require('path')
-var tokenizeArgString = require('./lib/tokenize-arg-string')
-var util = require('util')
+const camelCase = require('camelcase')
+const path = require('path')
+const statSync = require('fs').statSync
+const tokenizeArgString = require('./lib/tokenize-arg-string')
+const util = require('util')
```
evocateur: If consumers are webpacking yargs, that means we need to keep this lowest-common-denominator ES5-compatible, right? (no const, it would blow up UglifyJS :P)

PR #123, opened 2018-05-12
feat: add `set-placeholder-key` configuration

pvdlg:
If `true`, add a placeholder for each known key for which the corresponding CLI argument is not set.
The default value for the placeholder is `undefined`.

Follow up to https://github.com/yargs/yargs/pull/1105


pvdlg:
It seems Appveyor build fails because [it installs the latest version of npm](https://github.com/yargs/yargs-parser/blob/master/appveyor.yml#L9) on Node 5 which is not supported.

Issue #80, opened 2017-02-18
Array configured as camelCased option but value passed as hyphenated flag does't work


bcoe:
see: https://github.com/yargs/yargs/pull/770
see: https://github.com/yargs/yargs/issues/757

@laggingreflex I think we should be able to fix this issue in `yargs-parser` itself, which should have the advantage of fixing a few other classes of bugs; I hope to get to this today.

bcoe:
@laggingreflex I believe this should be fixed now?

PR #125, opened 2018-05-25
fix: #80 camelCase array options

laggingreflex:


bcoe:
@laggingreflex this should now be landed in `npm i yargs@next`, mind trying it out and letting me know if things work alright?

PR #130, opened 2018-08-18
Add halt-at-non-option configuration option

wmertens:
Fixes https://github.com/yargs/yargs/issues/946

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,25 +319,25 @@ node example.js -a 1 -c 2
 { _: [], a: 1, b: undefined, c: 2 }
 ```
 
-### stop at unknown
+### halt at non-option
 
 * default: `false`.
-* key: `stop-at-unknown`.
+* key: `halt-at-non-option`.
 
-Should parsing stop at the first unknown argument?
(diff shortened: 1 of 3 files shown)


wmertens:
@bcoe any chance of a quick merge? It's fully downwards-compatible…

bcoe:
@wmertens @Qard this feature feels a little bit magical to me. It's not really stopping on an unknown argument, it's essentially stopping parsing as soon as a positional argument is hit rather than a flag argument right?

In the cases where I've needed to split command line arguments (usually because one bin is instrumenting another bin) I've used a helper like this:

https://github.com/bcoe/c8/blob/master/lib/parse-args.js#L60

I wonder if instead of adding this additional parsing option, we could find a way to formalize this helper in yargs? I _think_ this is the use-case we're all trying to solve.

I'm swamped with open-source projects these days, so don't get to all my GitHub issues, perhaps I could convince you both to join me in slack here:

☝️ we've been using this slack to discuss JavaScript tooling related issues.

Issue #77, opened 2017-01-26
Unexpected behavior when combining dot-notation and camel-case expansion


adius:
`--dot-notation.foo.bar` yields to

```js
{
  'dot-notation': {foo: {bar: true}},
  dotNotationFooBar: true,
}
```

I would, however, expect it to get parsed to:

```js
{
  'dot-notation': {foo: {bar: true}},
  dotNotation: {foo: {bar: true}},
}
```

Is this a bug or the supposed behavior?

bcoe:
@adius this feels like a bug; if you get bored, would happily accept a patch 😉

iilei:
May I help with the Patch?

iilei:
Since the latest version this can be closed I suppose

bcoe:
@adius this should be fixed now thanks to @iilei's great work 👍

PR #131, opened 2018-08-22
Camel-Case-Expansion: alias by Property names, not Pathes

iilei:
This fixes issue #77
This *is* a breaking change

### DIFF
--- a/index.js
+++ b/index.js
@@ -371,7 +371,10 @@ function parse (args, opts) {
     unsetDefaulted(key)
 
     if (/-/.test(key) && configuration['camel-case-expansion']) {
-      addNewAlias(key, camelCase(key))
+      var alias = key.split('.').map(function (prop) {
+        return camelCase(prop)
+      }).join('.')
+      addNewAlias(key, alias)
     }
 
     var value = processValue(key, val)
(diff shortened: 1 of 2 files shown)


bcoe:
@iilei thank you for this contribution ... while this is a breaking change, it feels like it's definitely addressing a bug (and I can't imagine many folks in yargs relying on the old behavior). My temptation is to call it a breaking change in yargs-parser, but to make it a patch in yargs ... then there's a paper trail for folks if it does break anyone, but we don't release a new yargs major.

iilei:
@bcoe thank you for merging.

> My temptation is to call it a breaking change in yargs-parser, but to make it a patch in yargs

I would tend to do the same, for the same reasons. I can imagine it is more likely people have worked around it using aliases rather than relying on it. But that's just my uneducated guess

bcoe:
@iilei this is now available in `npm i yargs@next` mind taking it for a spin and letting me know if things work alright?

iilei:
@bcoe here is a code-kata repository / branch on which I verified it works:

https://github.com/iilei/kata-boilerplate/commit/0f244788b6eb246982123fbe9a8827323486ce13

PR #136, opened 2018-09-28
fix: make requiresArg work with arrays

vmx:
When `requiresArg` was used together with `array`, only the first
argument of the array would end up being parsed. With this change
things work as expected.

The `requireArg` option sets the `nargs` options to `1`, so only
one additional argument will be parsed. The problem was that that
parsing step was checked before the array parsing one. The solution
is as simple as flipping those if conditions and check if its an
array first, hence ignoring the `nargs` option.

Fixes https://github.com/yargs/yargs/issues/1170.

### DIFF
--- a/index.js
+++ b/index.js
@@ -166,12 +166,12 @@ function parse (args, opts) {
     )) {
       key = arg.match(/^--?(.+)/)[1]
 
-      // nargs format = '--foo a b c'
-      if (checkAllAliases(key, flags.nargs)) {
-        i = eatNargs(i, key, args)
       // array format = '--foo a b c'
-      } else if (checkAllAliases(key, flags.arrays) && args.length > i + 1) {
+      if (checkAllAliases(key, flags.arrays) && args.length > i + 1) {
         i = eatArray(i, key, args)
+      // nargs format = '--foo a b c'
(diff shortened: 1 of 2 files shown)


bcoe:
@vmx thank you for the contribution, I just restarted the build.

bcoe:
@vmx please try `npm i yargs@next` which should have this work landed \o/ if things are looking good, I'll land a new version of yargs early in the week.

vmx:
@bcoe I can confirm that with yargs@next (yargs@12.0.3-candidate.0) things now work as expected.

boneskull:
This change means I can't constrain the number of values read in an array-type option.  

IMO, this should be reverted until a better solution arises, because I cannot find a workaround.  Before this change, `check()` could have been used for validation.

bcoe:
@vmx, @boneskull and I just had a discussion around this here:

Because this is a breaking change, we're going to temporarily roll back for @boneskull's usage with mocha, but here's what I'd like to do:

1. we introduce the `greedy-arrays` option, which will default to `false` in `yargs-parser`; this will indicate that `array` array will parse a sequence of values like `-x 1 2 3` (behavior that can break commands, so should be opted into).
2. we'll re-and your work here; but make it so takes into account `greedy-arrays`.

PR #140, opened 2018-10-08
Ensure empty string is added into argv._

jacksdrobinson:
A string such as "hello '' world" would parse into ["hello", undefined, "world"].
This would add a catch for that, ensuring that an empty string is pushed
instead, and adds associated tests.

### DIFF
--- a/lib/tokenize-arg-string.js
+++ b/lib/tokenize-arg-string.js
@@ -17,6 +17,7 @@ module.exports = function (argString) {
     // split on spaces unless we're in quotes.
     if (c === ' ' && !opening) {
       if (!(prevC === ' ')) {
+        if (!args[i]) args[i] = ''
         i++
       }
       continue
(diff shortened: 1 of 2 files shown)


lib/tokenize-arg-string.js
```diff
@@ -17,6 +17,7 @@ module.exports = function (argString) {
     // split on spaces unless we're in quotes.
     if (c === ' ' && !opening) {
       if (!(prevC === ' ')) {
+        if (!args[i]) args[i] = ''
```
bcoe: :wave: for this fix, I think it would make sense to move the logic down to the check for the closing quote. So something like:

```
    if (c === opening) {
      if (!args[i]) args[i] = ''
      opening = null
      continue
  }
```

bcoe:
@jacksdrobinson thank you for the contribution, left a comment.

lib/tokenize-arg-string.js
jacksdrobinson: I see how that would make sense - placing it at the point of detecting the closing quotes. Both ways pass the tests, so I'll commit the change.

The reason I put it where I did was to ensure that we can't possibly increment `i` before we've checked for a missing argument.

jacksdrobinson: Committed :)

Issue #145, opened 2016-01-29
String arguments that contain dashes do not parse properly


andyburke:
Eg:

```
foo -a 'aaa' -b 'bbb' command "-- string with dashes in it --"
```

This will fall afoul of this case:

https://github.com/bcoe/yargs/blob/master/lib/parser.js#L103

I ran into this trying to encrypt a private ssh key using [cryptex](https://github.com/TechnologyAdvice/Cryptex).

bcoe:
@tristanMatthias @masaeedu I've transferred this issue to yargs-parser, would happily accept a patch if anyone feels like take a stab at the problem 👍

tristanMatthias:
@bcoe working on one now :)

tristanMatthias:
Ok @bcoe, I've submitted a pull request with the fix. Please note the solution required a little workaround (detailed in the PR) in which I add an extra set of quotation marks (then remove them later in the parser). The reason for this is because previously the tokenizer would create an array with `[...'--flag', '-valueStartingWithDash']`, and there was no way to determine if `'-valueStartingWithDash'` were flags or a string value (hence the extra wrapping). Let me know what you think!

masaeedu:
@bcoe I think the issue I'm having is a little different (maybe I need to open a different issue). When you pass something like the following: `node foo.js --hey "-what -is -going -on"`, this will be parsed by the shell into `['node', 'foo.js', '--hey', '-what -is -going -on']` and then passed into node. When yargs looks at this list, it should interpret the array element after `--hey` as a positional argument instead of as a flag, because it is meaningless as a flag. I think if the regexen in the parser were anchored to the end they wouldn't accept `-what -is -going -on` as a valid flag, and so it would be treated positional argument as desired.

As far as escaping quotes goes, we can double up quotes so they survive the shell's expansion process and make it into node's `argv`.

bcoe:
@masaeedu aha! I'd missed this edge-case, interesting.

@tristanMatthias thanks for a first pass at the solution, I wonder if we can figure out a way to avoid the temporary addition of quotes? working in the parser feels like a tech interview question doesn't it?

tristanMatthias:
It does indeed! Is there a way we can definitively differentiate between them with the quotes stripped? I thought about holding something in memory, but that doesn't work when it's passed from the Tokenizer to the Parser

masaeedu:
[...]

PR #146, opened 2018-11-09
Fix: 145

tristanMatthias:
This merge request fixes https://github.com/yargs/yargs-parser/issues/145, and adds the ability to pass in string values for flags or parameters that contain a `'-'` as the leading character. EG:

```js
const tokenize = require('./lib/tokenize-arg-string');
const parse = require('./');

const args = tokenize('--foo "-bar"');
// args is now [ '--foo', '"-bar"' ]. Notice the "..." (quotation marks) around the "-bar"
const parsed = parse(args);
// parsed is now { _: [], foo: '-bar' }
```

A risk with this is any value for a flag that has a value (as a string) starting with a dash is now returned with EXTRA quotation marks in the tokenized array. (See example above).

I've added a few extra test, and everything is running as normal, so it should be ok to merge in.

Let me know what you think!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,35 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="11.0.0"></a>
+# [11.0.0](https://github.com/yargs/yargs-parser/compare/v10.1.0...v11.0.0) (2018-10-06)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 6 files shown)


bcoe:
@masaeedu mind giving this a bit of a review, I don't have too many cycles to dig in for a few days, and want to make sure it matches your thinking around handling strings that contain `-`.

bcoe:
@tristanMatthias also, mind rebasing.

bcoe:
I used this as a jumping off point for https://github.com/yargs/yargs-parser/pull/153, which I'm closing this in favor of.

bcoe:
thank you for the contribution 👍

PR #150, opened 2018-11-19
Revert "fix: make requiresArg work in conjunction with arrays (#136)"

boneskull:
This reverts commit 77ae1d4e1c2590eeca025952671fff935ab7e884.

Prior to this commit, those wishing to require that an `array`-type option has a nonzero amount of arguments could use the `check()` function for validation.

The commit removed the ability to use `nargs` in conjunction with `array`-type options.

After this commit, those wishing to limit the number of values an `array`-type option accepts have no viable workaround.  This *also* has the implication that such an option cannot be used prior to a positional option, because it's ambiguous:

```bash
# usage: script [--foo n1 n2] <stuff>
$ script --foo bar baz quux
# `quux` should be `stuff`, but because `foo` is an array, it eats up `quux`.
```

**UPDATE**: Previous to #136, the above example was still an issue, but `nargs(1)` or `requireArg()` was a viable workaround; it no longer is.

### DIFF
--- a/index.js
+++ b/index.js
@@ -180,12 +180,12 @@ function parse (args, opts) {
     )) {
       key = arg.match(/^--?(.+)/)[1]
 
-      // array format = '--foo a b c'
-      if (checkAllAliases(key, flags.arrays) && args.length > i + 1) {
-        i = eatArray(i, key, args)
       // nargs format = '--foo a b c'
-      } else if (checkAllAliases(key, flags.nargs)) {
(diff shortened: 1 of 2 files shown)


boneskull:
@vmx looks like @bcoe will be landing this since it introduces problems.  I'll let him chime in, but you won't be left high & dry--it sounds like he'll have something to address your use case in the near future.

bcoe:
@vmx see my comments in #136, this is only a temporary revert.

PR #157, opened 2019-02-02
feat: no longer coerce numbers by default

bcoe:
This is a fairly big breaking change, but I think worthwhile. yargs and yargs-parser allow a user to opt in to numeric parsing on an option by option basis. Let's not magically convert strings to numbers by default, as this causes problems for phone #s, UUIDs, etc.

BREAKING CHANGE: folks must now opt in to parse-numbers, using configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ node example.js --foo.bar
 
 ### parse numbers
 
-* default: `true`
+* default: `false`
 * key: `parse-numbers`
 
 Should keys that look like numbers be treated as such?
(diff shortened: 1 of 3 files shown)


bcoe:
open question to me, is it worth forcing a large number of folks in the community to update their usage of the yargs/yargs-parser API, to remove this feature that I think is too magical and causes a variety of bugs.

bcoe:
after seeing the number of test failures in yargs from this change (not to mention the number of tests that needed to be updated in yargs-parser) opted for this less drastic change:

https://github.com/yargs/yargs-parser/pull/158

PR #163, opened 2019-03-22
Fix type "array" with "duplicate-arguments-array"

### LINKED ISSUE #162: type "array" with "duplicate-arguments-array" works inconsistently
`array` of type `number` is not affected by `duplicate-arguments-array` and doesn't reduce duplicated arguments:
```js
var args = parse(
    ['--file', '2', '--file', '6', '--timeout', '2s', '--timeout', '800'], {
        array: [{key: 'file', number: true}],
        string: ['timeout'],
        configuration: {
            'duplicate-arguments-array': false
        }
});
console.log(args);         // { _: [], file: [ 2, 6 ], timeout: '800' }
```
`array` of type `string` does reduce duplicated arguments with `duplicate-arguments-array`:
```js
var args = parse(
    ['--file', '2', '--file', '6', '--timeout', '2s', '--timeout', '800'], {
        array: [{key: 'file', string: true}],
        string: ['timeout'],
        configuration: {
            'duplicate-arguments-array': false
        }
});
console.log(args);         // { _: [], file: [ '6' ], timeout: '800' }
```
An `array` of type `boolean` is always reduced, indepently of ` duplicate-arguments-array`. So it doesn't seem possible to build an `array` of `boolean` with more than one element.

It is not clear:
- wether `duplicate-arguments-array` should reduce `array`'s at all. Analysing the tests, the answer is yes. Maybe it would be more logical to exclude `array`'s from this rule?
- the behavior of typed `array`'s is not consistent (number <=> string).

juergba:
### Description

closes #162 

The `array` type does not work consistently with option `duplicate-arguments-array`.
when `duplicate-arguments-array === false`
- `array` of type `string` **does** reduce duplicate arguments
- `array` of type `number` **does not** reduce duplicate arguments
- `array` of type `boolean` **only works** with arguments provided **in pairs**, eg `--foo true`.
A `boolean` without value (`--foo --no-foo`) results in various bugs as empty arrays, lost flags, lost array type and more.

### Description of Change
- `duplicate-arguments-array` has no effect on option type `array` anymore.
even when `duplicate-arguments-array === false`, you are able to specify:
(output depends on `flatten-duplicate-arrays`)
  - `-x 1 -x 2 -x 3`==> `[1, 2, 3]` or `[[1], [2], [3]]`
  - `-x 1 2 -x 4 5`==> `[1, 2, 4, 5]` or `[[1, 2], [4, 5]]`
- option type `number` and `string` work consistently.
- an `array` value is always an `array`, even if just one value is passed.
- `maybeCoerceNumber()`: fix handling of `array`s.
- adapt existing tests, add new tests

**Please note:** the disabling of creating an `array` by `-x a b c` by default will be implemented in another PR.

### DIFF
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -2200,25 +2200,33 @@ describe('yargs-parser', function () {
 
         parsed['x'].should.equal('b')
       })
-      it('array[string]: keeps only last argument', function () {
-        var parsed = parser('-x a -x b', {
-          array: [{ key: 'x', string: true }],
+      it('array[string]: keeps all arguments - ignores configuration', function () {
+        var parsed = parser('-x a -x b -y c', {
+          array: [
+            { key: 'x', string: true },
+            { key: 'y', string: true }
+          ],
           configuration: {
             'duplicate-arguments-array': false
           }
         })
 
-        parsed['x'].should.equal('b')
+        parsed['x'].should.deep.equal(['a', 'b'])
+        parsed['y'].should.deep.equal(['c'])
       })
-      it('array[number]: keeps only last argument', function () {
-        var parsed = parser('-x 1 -x 2', {
-          array: [{ key: 'x', number: true }],
+      it('array[number]: keeps all arguments - ignores configuration', function () {
+        var parsed = parser('-x 1 -x 2 -y 3', {
+          array: [
+            { key: 'x', number: true },
+            { key: 'y', number: true }
+          ],
           configuration: {
             'duplicate-arguments-array': false
           }
         })
 
-        parsed['x'].should.equal(2)
+        parsed['x'].should.deep.equal([1, 2])
+        parsed['y'].should.deep.equal([3])
       })
       it('does not interfere with nargs', function () {
         var parsed = parser('-x a b c -x o p q', {
@@ -2327,15 +2335,16 @@ describe('yargs-parser', function () {
             })
             parsed['x'].should.deep.equal([1, 2, 3])
           })
-          it('[-x 1 2 3 -x 2 3 4] => [2, 3, 4]', function () {
-            var parsed = parser('-x 1 2 3 -x 2 3 4', {
-              array: ['x'],
+          it('[-x 1 2 3 -x 2 3 4] => [[1, 2, 3], [2, 3, 4]]', function () {
+            var parsed = parser('-x 1 2 3 -x 2 3 4 -y 7', {
+              array: ['x', 'y'],
               configuration: {
                 'duplicate-arguments-array': false,
                 'flatten-duplicate-arrays': false
               }
             })
-            parsed['x'].should.deep.equal([2, 3, 4])
+            parsed['x'].should.deep.equal([[1, 2, 3], [2, 3, 4]])
+            parsed['y'].should.deep.equal([7])
           })
         })
         describe('type=number', function () {
@@ -2363,15 +2372,16 @@ describe('yargs-parser', function () {
             })
             parsed['x'].should.deep.equal([1, 2, 3])
           })
-          it('[-x 1 2 3 -x 2 3 4] => [2, 3, 4]', function () {
-            var parsed = parser('-x 1 2 3 -x 2 3 4', {
-              array: ['x'],
+          it('[-x 1 2 3 -x 2 3 4] => [1, 2, 3, 2, 3, 4]', function () {
+            var parsed = parser('-x 1 2 3 -x 2 3 4 -y 7', {
+              array: ['x', 'y'],
               configuration: {
                 'duplicate-arguments-array': false,
                 'flatten-duplicate-arrays': true
               }
             })
-            parsed['x'].should.deep.equal([2, 3, 4])
+            parsed['x'].should.deep.equal([1, 2, 3, 2, 3, 4])
+            parsed['y'].should.deep.equal([7])
           })
         })
         describe('type=number', function () {
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -2191,7 +2191,7 @@ describe('yargs-parser', function () {
 
         parsed['x'].should.deep.equal(['a', 'b'])
       })
-      it('keeps only last argument', function () {
+      it('string: keeps only last argument', function () {
         var parsed = parser('-x a -x b', {
           configuration: {
             'duplicate-arguments-array': false
@@ -2199,6 +2199,26 @@ describe('yargs-parser', function () {
         })
 
         parsed['x'].should.equal('b')
+      })
+      it('array[string]: keeps only last argument', function () {
+        var parsed = parser('-x a -x b', {
+          array: [{ key: 'x', string: true }],
+          configuration: {
+            'duplicate-arguments-array': false
+          }
+        })
+
+        parsed['x'].should.equal('b')
+      })
+      it('array[number]: keeps only last argument', function () {
+        var parsed = parser('-x 1 -x 2', {
+          array: [{ key: 'x', number: true }],
+          configuration: {
+            'duplicate-arguments-array': false
+          }
+        })
+
+        parsed['x'].should.equal(2)
       })
       it('does not interfere with nargs', function () {
         var parsed = parser('-x a b c -x o p q', {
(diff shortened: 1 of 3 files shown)


bcoe:
@juergba thank you for all the wonderful contributions you've made recently, I will work on getting them merged as soon as possible; I've been buried on other projects and apologize.

bcoe:
@juergba could I bother you to rebase? I'm getting ready to release a new version of yargs and yargs-parser.

test/yargs-parser.js
```diff
@@ -2200,6 +2200,26 @@ describe('yargs-parser', function () {
...
       })
+      it('array[string]: keeps only last argument', function () {
+        var parsed = parser('-x a -x b', {
+          configuration: {
+            array: [{ key: 'x', string: true }],
+            'duplicate-arguments-array': false
+          }
+        })
+
+        parsed['x'].should.equal('b')
```
bcoe: this behavior seems wrong to me, if we've explicitly indicated that `x` should be an array, the value should be `['b']` not `b` -- I also wonder if maybe `array` should take precedence over `duplicate-arguments-array`, and it should be `['a', 'b']` (I think this would be least shocking).

juergba: Yes, it's wrong, but I didn't open this coffin. The option type `string[]` behavior remains unchanged.

I tried to copy the option type `string[]` behavior to the option type `number[]`, to get consistent results.

juergba:
> [...] duplicate-arguments-array applying only to the behavior of keys you haven't explicitly set as an array.

Agree, type `array` should precede `duplicate-arguments-array`.
For this PR:
- behavior of `number[]` is correct
- behavior of `string[]` has to be fixed, should match `number[]` pattern
Note: this will probably "shock" many existing implementations / breaking change

> Honestly, I'd kind of like to drop support for -x a b c

What about `opts.narg` then?

coreyfarrell:
> my thinking is, if you've explicitly set type to `array`, we should still accept the option being set multiple times -- `duplicate-arguments-array` applying only to the behavior of keys you haven't explicitly set as an array.

Yes please.  In the context of nyc I would like to block use of `--temp-dir` or `--cwd` multiple times but not block multiple use of use of `--reporter`.

> Honestly, I'd kind of like to drop support for `-x a b c`, and make it so the only way to create an array is `-x a -x b -x c`; since type `array` does not mix well with positional arguments.

Maybe it would be better to disable `-x a b c` by default but still have a way to enable it?  Worth considering that `-x a b c -x d e f` can mean `x: [['a', 'b', 'c'], ['d', 'e', 'f']]`?

bcoe:
> what about opts.narg then?

I think `narg` would behave differently than `array` and just take the last set. so, if `narg` of `x` was set to `2`, and you passed `--x a b --x c d`, you'd get `c`, `d`.

@coreyfarrell 

> Maybe it would be better to disable -x a b c by default but still have a way to enable it? 

I agree with this, maybe a parameter called `--positional-arrays`, which we start to default to `false`? Then the `array` type can be used (by default) simply to imply:

1. this value will always be an array, even if just one value is passed.
2. even if you set `duplicate-arguments-array` to `false`, you will still be able to specify `-x 1 -x 2 -x 3`.

@juergba does this sound reasonable to you for mocha's needs? If so perhaps we could update this patch to reflect this world.

juergba:
> does this sound reasonable to you for mocha's needs?

afaik we don't use `-x a b c` in Mocha, we only use `-x a -x b -x c` or `-x a,b,c`.

> If so perhaps we could update this patch to reflect this world.

Ok, I will give it a try. I'm not sure yet about the implications of "disabling `-x a b c` by default".
I will see ...

test/yargs-parser.js
juergba: done

juergba:
@bcoe  The behavior of option type `boolean[]` seems to be completely buggy, even before this PR.
- are option type `boolean[]` allowed?
- is their expected behavior identical to `number[]` and `string[]`?

bcoe:
@juergba my thinking is we should make this a breaking change, and address some of these long standing issues. If you indicate something is an array, even a `boolean`, I think you should get an array out the other end, even if it's an empty array (if no argument is set) or an array of one item if only one item is provided.

It might break the internet too much to drop support for `-x a b c`, but perhaps we could at least add a config setting that allows this to be disabled? Then consider whether we drop the feature eventually, forcing `-x a -x b -x c`.

juergba:
I understood your concerns with `-x a b c` and your plans for an additional config setting. IMO before doing this we should fix some bugs.

The current master shows following results:
```js
  'array': ['file'],
  'number': ['file'],           // { _: [], file: [ 0, 0, undefined ] }
  // 'string': ['file'],        // { _: [], file: [ [], [], '' ] }
  // 'boolean': ['file'],       // { _: [], file: [ [], [], [] ] }
  configuration: {
  }
})
```
- two flags => three values parsed
- `number` and `string` show default value, `boolean` is empty

```js
var args = parse('--file true --file true', {   // same result with '--file --file'
  'boolean': ['file'],    
  configuration: {
    'duplicate-arguments-array': true
  }
})           // { _: [], file: true }
```
- result should be: `[true, true]`, but is `true`

We should address those bugs first, before adding new flags. `boolean` is a special option since the flag itself can carry its value, so `--booly` === `--booly true`.

bcoe:
@juergba I agree, let's start by fixing these bugs.

Do we want to close this issue, in favor of landing fixes like #184? and moving towards a better API for arrays in the long run?

juergba:
> Do we want to close this issue

No, please don't close. This issue covers its own topic, which got lost completely during  `-x a b c` discussion.

> [...] moving towards a better API for arrays in the long run?

Yes, landing fixes step by step is a good procedure.

bcoe:
@juergba any chance we can dust this PR off, I'd like to start working on getting a few more releases of yargs out the door.

juergba:
Yes, I can make the original issue (type "array" with "duplicate-arguments-array") ready within a few days.

The disabling of creating an array by `-x a b c` by default is not part of this PR.

bcoe:
@juergba sounds good to me, I think that's @coreyfarrell's recommendation to make disable `-x a b c` by default might be smart honestly, but in a future PR ... my concern is this would likely bite quite a few folks, so it's worth opening a discussion.

juergba:
> my thinking is, if you've explicitly set type to array, we should still accept the option being set multiple times -- duplicate-arguments-array applying only to the behavior of keys you haven't explicitly set as an array.

@bcoe please confirm wether you really want above pattern:
 (`duplicate-arguments-array: false`)
`[-x 1 2 3 -x 2 3 4] => [[1, 2, 3], [2, 3, 4]]`

In #75 the opposite was implemented:
`[-x 1 2 3 -x 2 3 4] => [2, 3, 4]`

bcoe:
@juergba if you have `duplicate-arguments-array = true`, and an option is an `array`, I feel it should be:

`[[1, 2, 3], [2, 3, 4]]`

If you have `duplicate-arguments-array = true`, and an option is an `array` I feel it should be:

`[2, 3, 4]`.

This agrees with your thinking?

juergba:
No, this is not what I'm thinking.

Neither after your comment:
> my thinking is, if you've explicitly set type to array, we should still accept the option being set multiple times -- duplicate-arguments-array applying only to the behavior of keys you haven't explicitly set as an array.

Nor after @coreyfarrell 's comment:
> Yes please. In the context of nyc I would like to block use of --temp-dir or --cwd multiple times but not block multiple use of use of --reporter.

@coreyfarrell can you help, please?

bcoe:
@juergba `duplicate-arguments-array` is a specific setting which indicates that the last time a value is set, like `-x` it should take preference.

Therefore, in the edge case that you've set a `array` and `duplicate-arguments-array` it should have the same behavior:

`-x 3 5 5 -x 4 3 2` in which the last `-x` is preferred, so:

`x: [4, 3, 2]`.

I don't think we should change this behavior ... and it's pretty edge-casey to be honest (not something I've seen many issues about).

What I would be more willing to entertain would be a redefinition of what setting the type `array` means, i.e., getting rid of support for `-x 3 2 4` entirely.

tldr; I'm not super excited about the changes proposed in this PR because it seems like thrash without solving much of a problem I've observed, I'd rather work to make the `array` setting less surprising (and perhaps retire some of the config settings around arrays in this process).

bcoe:
getting back to your original PR description:

## If duplicate-arguments-array = false

* `-x a b c -x d e f` and `x = string`, `array = true`, should result in `[d, e, f]`
* `-x 1 2 3 -x 4 5 6` and `x = number`, `array = true`, should result in `[4, 5, 6]`.
* `-x a b c -x d e f` and `x = string`, `array = false`, should result in `d`.
* `-x 1 2 3 -x 4 5 6` and `x = number`, `array = false`, should result in `4`.

## If duplicate arguments-array = true

* `-x a b c -x d e f` and `x = string`, `array = true`, should result in `[[a, b, c], [d, e, f]]`
* `-x 1 2 3 -x 4 5 6` and `x = number`, `array = true`, should result in `[[1, 2, 3], [4, 5, 6]]`.
* `-x a b c -x d e f` and `x = string`, `array = false`, should result in `[a, d]`.
* `-x 1 2 3 -x 4 5 6` and `x = number`, `array = false`, should result in `[1, 4]`.
* `-x a -x b -x c` and `x = string`, `array = false`, should result in `[a, b, c]`.
* `-x 1 -x 2 -x 3` and `x = number`, `array = false`, should result in `[1, 2, 3]`.

----

:point_up: if there are any fixes we can make around these expectations let's make them :ok_hand: 

And then I'd say, revisit the meaning of `array` in a separate issue (as I think it's confusing).

juergba:
Meanwhile a few PR's refering `boolean` and `array` have been merged. This PR can be closed.

bcoe:
@juergba :+1: sorry for this dragging for so long, FYI any chance I could get you to join the slack here:

I'd like to roll out a new yargs with a bunch of your fixes, but would first like to make sure that it seems to behave nicely with `mocha`.

PR #172, opened 2019-04-04
feat: Add `strip-aliased` and `strip-dashed` configuration options.

coreyfarrell:
Enabling these options cause redundant keys to be removed from the
results.  This can be useful when printing the results for debugging as
it reduces the amount of information to look at.


bcoe:
@coreyfarrell these contributions you've been making are great, sorry for sitting on this for so long -- have been buried jumping between other repos.

PR #181, opened 2019-06-07
feature: add configuration option "collect-unknown-options"

henderea:
Fixes yargs/yargs#1243

All new code is covered in tests. It shouldn't make any difference, but I used the Wallaby.js VSCode extension for test running and coverage, instead of the npm scripts.

I tried to match the existing code style as well as possible, but please feel free to fix any issues I created.

### DIFF
--- a/index.js
+++ b/index.js
@@ -220,7 +220,6 @@ function parse (args, opts) {
     } else if (arg.match(/^-[^-]+/) && !arg.match(negative)) {
       letters = arg.slice(1, -1).split('')
       broken = false
-      let unmatched = []
 
       for (var j = 0; j < letters.length; j++) {
         next = arg.slice(j + 2)
@@ -293,9 +292,6 @@ function parse (args, opts) {
             setArg(key, defaultValue(key))
           }
         }
-      }
-      if (unmatched.length > 0) {
-        argv._.push(['-', ...unmatched].join(''))
       }
     } else if (arg === '--') {
       notFlags = args.slice(i + 1)
--- a/index.js
+++ b/index.js
@@ -26,7 +26,8 @@ function parse (args, opts) {
     'set-placeholder-key': false,
     'halt-at-non-option': false,
     'strip-aliased': false,
-    'strip-dashed': false
+    'strip-dashed': false,
+    'ignore-unknown-options': false
   }, opts.configuration)
   var defaults = opts.default || {}
   var configObjects = opts.configObjects || []
@@ -159,12 +160,18 @@ function parse (args, opts) {
       } else if (checkAllAliases(m[1], flags.arrays)) {
         args.splice(i + 1, 0, m[2])
         i = eatArray(i, m[1], args)
-      } else {
+      } else if (!configuration['ignore-unknown-options'] || hasAnyFlag(m[1])) {
         setArg(m[1], m[2])
+      } else {
+        argv._.push(maybeCoerceNumber('_', arg))
       }
     } else if (arg.match(negatedBoolean) && configuration['boolean-negation']) {
       key = arg.match(negatedBoolean)[1]
-      setArg(key, checkAllAliases(key, flags.arrays) ? [false] : false)
+      if (!configuration['ignore-unknown-options'] || hasAnyFlag(key)) {
+        setArg(key, checkAllAliases(key, flags.arrays) ? [false] : false)
+      } else {
+        argv._.push(maybeCoerceNumber('_', arg))
+      }
 
     // -- separated by space.
     } else if (arg.match(/^--.+/) || (
@@ -179,7 +186,7 @@ function parse (args, opts) {
       // array format = '--foo a b c'
       } else if (checkAllAliases(key, flags.arrays)) {
         i = eatArray(i, key, args)
-      } else {
+      } else if (!configuration['ignore-unknown-options'] || hasAnyFlag(key)) {
         next = args[i + 1]
 
         if (next !== undefined && (!next.match(/^-/) ||
@@ -194,29 +201,40 @@ function parse (args, opts) {
         } else {
           setArg(key, defaultValue(key))
         }
+      } else {
+        argv._.push(maybeCoerceNumber('_', arg))
       }
 
     // dot-notation flag separated by '='.
     } else if (arg.match(/^-.\..+=/)) {
       m = arg.match(/^-([^=]+)=([\s\S]*)$/)
-      setArg(m[1], m[2])
+      if (!configuration['ignore-unknown-options'] || hasAnyFlag(m[1])) {
+        setArg(m[1], m[2])
+      } else {
+        argv._.push(maybeCoerceNumber('_', arg))
+      }
 
     // dot-notation flag separated by space.
     } else if (arg.match(/^-.\..+/)) {
       next = args[i + 1]
       key = arg.match(/^-(.\..+)/)[1]
 
-      if (next !== undefined && !next.match(/^-/) &&
-        !checkAllAliases(key, flags.bools) &&
-        !checkAllAliases(key, flags.counts)) {
-        setArg(key, next)
-        i++
+      if(!configuration['ignore-unknown-options'] || hasAnyFlag(key)) {
+        if (next !== undefined && !next.match(/^-/) &&
+          !checkAllAliases(key, flags.bools) &&
+          !checkAllAliases(key, flags.counts)) {
+          setArg(key, next)
+          i++
+        } else {
+          setArg(key, defaultValue(key))
+        }
       } else {
-        setArg(key, defaultValue(key))
+        argv._.push(maybeCoerceNumber('_', arg))
       }
     } else if (arg.match(/^-[^-]+/) && !arg.match(negative)) {
       letters = arg.slice(1, -1).split('')
       broken = false
+      unmatched = []
 
       for (var j = 0; j < letters.length; j++) {
         next = arg.slice(j + 2)
@@ -233,8 +251,12 @@ function parse (args, opts) {
           } else if (checkAllAliases(key, flags.arrays)) {
             args.splice(i + 1, 0, value)
             i = eatArray(i, key, args)
-          } else {
+          } else if (!configuration['ignore-unknown-options'] || hasAnyFlag(key)) {
             setArg(key, value)
+          } else {
+            unmatched.push(key);
+            unmatched.push('=');
+            unmatched.push(value);
           }
 
           broken = true
@@ -242,24 +264,41 @@ function parse (args, opts) {
         }
 
         if (next === '-') {
-          setArg(letters[j], next)
+          if (!configuration['ignore-unknown-options'] || hasAnyFlag(letters[j])) {
+            setArg(letters[j], next)
+          } else {
+            unmatched.push(letters[j]);
+            unmatched.push(next);
+          }
           continue
         }
 
         // current letter is an alphabetic character and next value is a number
         if (/[A-Za-z]/.test(letters[j]) &&
           /^-?\d+(\.\d*)?(e-?\d+)?$/.test(next)) {
-          setArg(letters[j], next)
+          if (!configuration['ignore-unknown-options'] || hasAnyFlag(letters[j])) {
+            setArg(letters[j], next)
+          } else {
+            unmatched.push(letters[j]);
+            unmatched.push(next);
+          }
           broken = true
           break
         }
 
         if (letters[j + 1] && letters[j + 1].match(/\W/)) {
-          setArg(letters[j], next)
+          if (!configuration['ignore-unknown-options'] || hasAnyFlag(letters[j])) {
+            setArg(letters[j], next)
+          } else {
+            unmatched.push(letters[j]);
+            unmatched.push(next);
+          }
           broken = true
           break
-        } else {
+        } else if (!configuration['ignore-unknown-options'] || hasAnyFlag(letters[j])) {
           setArg(letters[j], defaultValue(letters[j]))
+        } else {
+          unmatched.push(letters[j])
         }
       }
 
@@ -273,7 +312,7 @@ function parse (args, opts) {
         // array format = '-f a b c'
         } else if (checkAllAliases(key, flags.arrays)) {
           i = eatArray(i, key, args)
-        } else {
+        } else if (!configuration['ignore-unknown-options'] || hasAnyFlag(key)) {
           next = args[i + 1]
 
           if (next !== undefined && (!/^(-|--)[^-]/.test(next) ||
@@ -288,7 +327,12 @@ function parse (args, opts) {
           } else {
             setArg(key, defaultValue(key))
           }
+        } else {
+          unmatched.push(key);
         }
+      }
+      if (unmatched.length > 0) {
+        argv._.push(maybeCoerceNumber('_', ['-', ...unmatched].join('')))
       }
     } else if (arg === '--') {
       notFlags = args.slice(i + 1)
@@ -752,6 +796,17 @@ function parse (args, opts) {
 
     toCheck.forEach(function (key) {
       if (flag.hasOwnProperty(key)) isSet = flag[key]
+    })
+
+    return isSet
+  }
+  
+  function hasAnyFlag (key) {
+    var isSet = false
+    var toCheck = [].concat(Object.values(flags))
+
+    toCheck.forEach(function (flag) {
+      if (flag[key]) isSet = flag[key]
     })
 
     return isSet
--- a/index.js
+++ b/index.js
@@ -27,7 +27,7 @@ function parse (args, opts) {
     'halt-at-non-option': false,
     'strip-aliased': false,
     'strip-dashed': false,
-    'ignore-unknown-options': false
+    'parse-unknown-options': true
   }, opts.configuration)
   var defaults = opts.default || {}
   var configObjects = opts.configObjects || []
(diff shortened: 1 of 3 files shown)


bcoe:
@henderea thanks for the patch. Mind running `./node_modules/.bin/standard --fix`. 

@juergba bother you for a code review?

juergba:
@bcoe I will need some days for my code review.

On my first glance the documentation is missing.

> _: an array representing the positional arguments.

I'm not sure wether it is a good idea to use property `_` for this purpose. Positional arguments and unkown options are completely different objects.

henderea:
@juergba I can work on documentation, but the whole point of this feature is to treat unknown options like regular arguments.  If it doesn't do that, then the whole change is pointless for my needs.

Perhaps the name should be updated?  The idea of this mode is to treat unknown options like regular arguments, so I'm pretty sure `_` is the right place for that.  I'm not always the best at naming in code 😅

If you look at the issue I filed that this PR fixes, yargs/yargs#1243, you'll see that my intention is to allow unknown options to appear in a varargs of a command, so unless `_` won't do that, I'm pretty sure I'm putting it in the right place.

index.js
```diff
@@ -295,8 +334,13 @@ function parse (args, opts) {
           } else {
             setArg(key, defaultValue(key))
           }
+        } else {
+          unmatched.push(key)
         }
       }
+      if (unmatched.length > 0) {
+        argv._.push(maybeCoerceNumber('_', ['-', ...unmatched].join('')))
```
juergba: the values are already processed, right?

index.js
```diff
@@ -241,33 +259,54 @@ function parse (args, opts) {
           } else if (checkAllAliases(key, flags.arrays) && args.length > i + 1) {
             args.splice(i + 1, 0, value)
             i = eatArray(i, key, args)
-          } else {
+          } else if (!configuration['ignore-unknown-options'] || hasAnyFlag(key)) {
             setArg(key, value)
+          } else {
+            unmatched.push(key)
+            unmatched.push('=')
```
juergba: why with "="?

henderea: Otherwise the `=` is lost, that's why.

index.js
```diff
@@ -168,12 +169,18 @@ function parse (args, opts) {
       } else if (checkAllAliases(m[1], flags.arrays) && args.length > i + 1) {
         args.splice(i + 1, 0, m[2])
         i = eatArray(i, m[1], args)
-      } else {
+      } else if (!configuration['ignore-unknown-options'] || hasAnyFlag(m[1])) {
         setArg(m[1], m[2])
+      } else {
+        argv._.push(maybeCoerceNumber('_', arg))
```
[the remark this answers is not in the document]
henderea: Good point, I'll remove the number processing.

index.js
henderea: Fair point.  I'll remove the number parsing here too.

index.js
```diff
@@ -26,7 +26,8 @@ function parse (args, opts) {
     'set-placeholder-key': false,
     'halt-at-non-option': false,
     'strip-aliased': false,
-    'strip-dashed': false
+    'strip-dashed': false,
+    'ignore-unknown-options': false
```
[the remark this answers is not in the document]
henderea: I'm definitely open to naming suggestions.  If you think `parse-unknown-options` is clearer, then I'd be happy to go with that.  If you have other suggestions, I'm open to those too.

henderea:
Node 6 support fixed (didn't realize `Object.values` was an issue, probably should have looked at the build failure 😅).  I've also changed it to `parse-unknown-options: false`, and removed the `maybeCoerceNumber` from my new code.  It seems the existing number coercion for unhandled args in the final else is what caused the numeric arguments in my tests, so my tests were actually not changed by the removal of coercion from my new code.

I'll plan to expand tests over the weekend if I get to it.

bcoe:
@henderea @juergba perhaps we can take a step back here, why do you want this behavior?

Tools like [nyc](https://github.com/istanbuljs/nyc) and [mocha](https://github.com/mochajs/mocha) have similar needs, in which they want to be able to run a command where some options are intended for the application, and some are intended for the program being instrumented, e.g.,

```
nyc --option 1 foo --option 3 --option 4
```

We've introduced the configuration [halt-on-non-option(https://github.com/yargs/yargs-parser#halt-at-non-option) to support this use-case, which I think is pretty close to what you're trying to do?

bcoe:
@henderea just bumping up this conversation,  was wondering if `--halt-on-non-option` could potentially give you close to the behavior that you're looking for, for the application you're building.

The behavior of stopping parsing once you hit the first non `--option` argument is fairly common for tools that instrument a file (wondering if this is close to your use-case).

henderea:
@bcoe: sorry, haven't gotten around to expanding the test suite yet

As for `--halt-on-non-option`, my use case is actually related to passing a full command string (including flags) as a parameter to a script.  Basically, i want to be able to have `./script.js command ls -lh ~` (with command defined as `'command <exe> [args...]'`) be able to capture `-lh ~` in the `args`.

Does `--halt-on-non-option` allow flags to be captured in a varargs when appearing after a non-option?  if so, then that might do what I need.  But if the `'command <exe> [args...]'` command definition won't allow me to capture options untouched in the `args` when `<exe>` is a non-option, then it won't work for my needs.

henderea:
Also, related to `--halt-on-non-option`, can that be applied only for certain commands?  I would still like other commands to be able to end with a flag, placed after regular arguments.

The tool I want this for is something I'm converting from ruby, and the ruby library I use for command and option handling by default behaves like my `parse-unknown-options: false` code, treating unknown options as regular positional arguments.

I'll try to get some additional tests written, but I think my needs will only be fully handled by something similar to this pull request.

henderea:
For example, I want to have the following definitions:

`'list <pattern>'` with boolean option `-l`
and
`'run <exe> [args...]'` with no options

I want to be able to run
`./script.js list -l pat`
and
`./script.js list pat -l`
with both setting
```
{
  pattern: 'pat',
  l: true
}
```
but have
`./script.js run ls -l .`
set
```
{
  exe: 'ls',
  args: ['-l', '.']
}
```

As a note, the `list` command lists something from a specific JSON file, and `run` runs an arbitrary shell command.  There is nothing in common between `list` and `run ls`.  Also note that this is a simplified set of commands with different names from my actual intended usage, but they _are_ representative of my desired usage and behavior.

henderea:
@bcoe: If possible, I would like for this feature or something else that will treat unknown options as regular arguments to be added to yargs.  It looks like the `halt-on-non-option` flag would have to apply to _all_ commands, but I want to parse options both before _and_ after positional arguments on commands that have options.  If I don't want to parse any options on a given command, I will just not define any.

When I write a script, I will define EVERY option that I want to handle, and anything I _don't_ define shouldn't be treated as an option.  The ruby library I use for argument parsing (thor) does this _by default_, and I personally feel that it is rather odd to set options that haven't been defined.  The way I see it, my script should only support options that have been defined, and should fail on undefined options, _but only if they don't fit into an argument or varargs_.

bcoe:
@henderea :wave: thanks for filling me in regarding your use case, I'll try to get back to you with a response soonish :smile: when I can give this my full attention.

index.js
```diff
@@ -751,6 +795,18 @@ function parse (args, opts) {
     return isSet
   }
 
+  function hasAnyFlag (key) {
+    var isSet = false
+    // XXX Switch to [].concat(...Object.values(flags)) once node.js 6 is dropped
```
bcoe: I like the method name `hasUnknownOption`, which would take an `arg` and `arg + 1` and:

1. if argument is form `-foo`, and `arg + 1` starts with `-`, it would ensure that `f`, `o`, and `o`, are known keys.
2. if argument is form `--foo`, or `--foo=bar`, it would check for `foo`.

Rather than using the approach:

```
var toCheck = [].concat(...Object.keys(flags).map(k => flags[k]))
```

I think you'd be best to check `opts.key`, which will be set even if you only call `yargs.option('foo', {describe: 'foo key'})`.

☝️ I think if you use this approach you can just call:

```
checkAllAliases(key, opts.key)
```

juergba:
> I agree with @juergba, I would introduce a new key on argv called unknownOptions, which for your use case could just be combined with _ in your application.

Meanwhile I have changed my mind on this:
- `halt-at-non-option` also uses the `_` property, so unparsed arguments are already treated as positional arguments.
- let's take follwing example: "--ls aa bb cc". `ls` is an unknown option, so we don't know anything about its type and how many of the following arguments should have to be consumed.
  - should the values `aa`, `bb`, `cc` be handled as positional arguments ==>  `arg._` ?
  - should we handle them as input to the unknown option `ls` ==> `argv.unknownOptions` ?

We could end up with spreading our unknown arguments over to different properties.

> ☝️ hasUnknownOption would just need to drop handle:
-xyz: ensuring that x, y, and z are all configured.
--foo=bar: extracting just the foo and checking that it's configured.
--foo: extracting just the foo and making sure it's configured.

This list is not complete, the extraction of the plain key out of the raw argument is more complex than this. E.g. boolean negation, dot anotation are missing. We would have to copy the extraction logic into `hasUnknownOption`, I'm not convinced of this.

bcoe:
@juergba I'd like to avoid adding conditional logic throughout the parser, to handle the `--unknown-options` case, I'd also like to avoid logic that requires that a key be rebuilt with less than perfect information, e.g, the case of:

```js
_.push(key)
_.push('=')
```

we could just continue to enumerate some of these other cases:

```
--no-[key]
--[key].foo
```

☝️ it's a finite list.

henderea:
Well, I don't necessarily have an issue with simplifying the logic if it still does the same thing, but putting the unknown options into a separate property and merging them could alter the order of arguments, which would not be desirable.

If there is a way to rewrite the logic to get the same effect with less-invasive changes, I'm all for that.  I just want to be sure it still works the same way, or at least that I can get the same result through some method.  Altering the order of arguments because they were split between 2 different properties is not desireable.

bcoe:
@henderea I'm open to putting the options in `_` :+1: but I like the idea of encapsulating the logic for skipping a key in the first check of our `if`/`else` block, this way we don't end up breaking the feature over and over again as we make slight tweaks to different option types (I've seen this happen for other constructs like negation and dot notation).

bcoe:
@henderea just following up 👍 to summarize:

1. I'm comfortable with the idea of us putting the unknown options in `_`.
2. I'd like the check for an unknown option to ideally be its own logic before the rest of option parsing, so that we don't have to perform the check in a variety of places (even though it creates some extra parsing logic in the case of `collect-unknown-options`, I like that we don't need to add additional checks to ever other type of argument).
3. let's use `collect-unknown-options` as the config setting.

henderea:
@bcoe Sure, I can make the changes.  The thing is, I'm not 100% sure of the form those changes would take.  How would I detect that a key should be skipped before parsing the actual key?  If you could give me a bit more direction on that, I could make the changes.

bcoe:
@henderea I can try to pseudo code something, but my thinking was you only have a few forms of keys:

```
--foo
-foo
--foo=99
-foo=99
```

☝️ so have a little helper function that runs before the rest of our checks and extracts the `foo`, if we handle `foo` in our parser, move to the next steps, otherwise put the whole key `--foo`, or `--foo=99`, etc., into unknown arguments.

henderea:
@bcoe I've made the requested changes.  I ended up creating a function that takes `arg` and some regex patterns, and checks if parsing `arg` with any of the patterns results in a key (in match group 1) that has been configured.  The name I gave it is `checkFlags`, but you may want to give it a better name.  Code naming isn't a strong point of mine 😅

henderea:
I realized after my first implementation of the updated code that it wasn't checking short flags individually, so I updated my code to handle that, and also added in the check for negative numbers that I saw in the short flags handling, along with renaming `checkFlags` to `hasFlagsMatching` so it matches more closely with my `hasAnyFlag` and new `hasAllShortFlags` functions.

bcoe:
@henderea thanks for taking this on :+1: will provide review soon.

index.js
```diff
@@ -142,8 +143,10 @@ function parse (args, opts) {
     var next
     var value
 
+    if (configuration['collect-unknown-options'] && !arg.match(negative) && !hasAllShortFlags(arg) && !hasFlagsMatching(arg, /^-+([^=]+?)=[\s\S]*$/, /^-+no-([^=]+?)=[\s\S]*$/, /^-+([^=]+?)$/, /^-+no-([^=]+?)$/, /^-+([^=]+?)-$/, /^-+([^=]+?)\d+$/, /^-+([^=]+?)\W+.*$/)) {
```
bcoe: let's pull the bits after the `&&` into a helper called `isUnknownOption`; these regexes are making my eyes bleed, and will read better if they're assigned to variables in a helper function that describe their purpose.

henderea: @bcoe I've implemented this requested change.  I also added some comments in the `isUnknownOption` and `hasAllShortFlags` explaining the logic a bit.

henderea:
It looks like the windows builds have some issues on travis:

```
$ nvs add 10
Downloading bootstrap node from https://nodejs.org/dist/v10.12.0/node-v10.12.0-win-x64.7z
######################################################################## 100.0%
The command "nvs add 10" failed and exited with 1 during .
```

I think I saw this same type of error for an `nvs add 8` command on a windows node 8 build for this PR before.

On that note, is there a good way to re-trigger the build?

henderea:
@bcoe if the code is good, is that it for my part of the PR?  Or do I need to do some sort of changelog entry or other non-code stuff like I did when I contributed to RVM once?

index.js
```diff
@@ -807,6 +809,25 @@ function parse (args, opts) {
     return hasAllFlags
   }
 
+  function isUnknownOption (arg) {
```
bcoe: this refactor makes me quite happy :smile:

bcoe:
@henderea all the CHANGELOG updates, etc., are automatic :+1:

We should however update the README.md with the new feature. Haven't forgotten about you, have just been overcommitted on a few other projects. Your work is appreciated.

henderea:
@bcoe I guess I was assuming the README.md changes were something you would handle and that you were just mentioning it, but did you actually mean for _me_ to handle it?  I can do it if needed, I just want to make sure you aren't waiting on me to do something I wasn't aware was on my plate 😅

bcoe:
@henderea mind adding a new section to the README describing the new configuration setting, I can provide review on the wording afterwards :+1: 

If you don't have the time, I can try to write something up when I land this.

henderea:
@bcoe: updated the readme.  here's what i put (based off of "strip dashed" and put right after it, at the end of the section):

### collect unknown options

* default: `false`
* key: `collect-unknown-options`

Should unknown options be collected into `_`?  An unknown option is one that is not configured in `opts`.

_If disabled_

```sh
node example.js --unknown-option --known-option 2
{ _: [], unknownOption: true, knownOption: 2 }
```

_If enabled_

```sh
node example.js --unknown-option --known-option 2
{ _: ['--unknown-option'], knownOption: 2 }
```

kellyselden:
This feature worked as needed for my use case. Given code like:

```
  let parsed = parse(process.argv.slice(3), {
    string: ['b'],
    configuration: {
      'collect-unknown-options': true
    }
  });
```

(slice(3) since it is a command) yields

```
Object {_: Array(2), b: "foo"}
    _:Array(2) ["--unused-option1=3", "--unused-option2=5"]
    b:"foo"
    __proto__:Object {constructor: , __defineGetter__: , __defineSetter__: , …}
```

bcoe:
@henderea could I bother you to try:

```
npm i yargs@next
```

It has your changes 🎉 thanks for being patient.

kellyselden:
I just tried this out, and it's working unexpectedly for me.

Screenshot 2019-09-07 at 09 49 21

I'm inside the `init` command `handler`. I'm testing it via vscode with

```
    {
      "type": "node",
      "request": "launch",
      "name": "Launch Program",
      "program": "${workspaceFolder}/bin/ember-cli-update.js",
      "args": [
        "init",
        "-b=temp",
        "--custom"
      ]
    },
```

Where `init` is the expected command, `-b` is an expected option, and `--custom` is the unexpected option. `argv._` in this case has the unexpected option and excludes the expected option correctly, but it also includes the command in the list. This doesn't seem right. Am I doing this correctly?

kellyselden:
Here is the PR for more context https://github.com/ember-cli/ember-cli-update/pull/688

bcoe:
What's the result in yargs@14.0.0, my expectation as well is that "init" wouldn't be populated.

kellyselden:
Here it is back to yargs v14.0.0

Untitled

henderea:
@bcoe doesn't work for me either

**index.js**
```javascript
const argv = yargs
    .parserConfiguration({
        'collect-unknown-options': true
    })
    .usage('Usage: $0 command [args]')
    .wrap(120)
    .command(require('../lib/commands/sudo'))
    .argv
```

**lib/commands/sudo.js**
```javascript
exports.command = ['sudo <command> [args..]', 's'];
exports.describe = 'run a sudo with the stored password filled in for you';

exports.builder = yargs => {
    return yargs.positional('command', {
        describe: 'The executable to run'
    }).positional('args', {
        describe: 'The arguments to the executable'
    }).option('rvm-sudo', {
        describe: 'use rvmsudo instead of sudo',
        type: 'boolean'
    });
};

exports.handler = async argv => {
    console.log(argv);
};
```

**output (14.1.0)**
```
$ node ./dist/index.js sudo ls -l
{ _: [ 'sudo' ], args: [ [] ], '$0': 'dist/index.js', command: 'ls' }
```

**output (14.0.0)**
```
$ node ./dist/index.js sudo ls -l
{
  _: [ 'sudo' ],
  l: true,
  args: [],
  '$0': 'dist/index.js',
  command: 'ls'
}
```

`-l` should appear in the `args` array in 14.1.0, but it just disappears completely.  Does the command parsing not use `_` from the parser internally?

bcoe:
@henderea command parsing does partially use yargs-parser directly, see:

henderea:
Yep, I see this in the code:
https://github.com/yargs/yargs/blob/master/lib/command.js#L339
```delete options.config```

That's where the issue is.  It needs to keep the new config option I created.

bcoe:
@henderea cool :+1: if you don't mind taking a look, I can look at the other issue of the `command` still being populated ... I think that might have honestly been intentional, to provide context to the handler about what command invoked it ... this at least rings a bell for me.

henderea:
@bcoe: created #202 since I can't re-open this PR.  I found that the reason it was still not working like I expected was because it ran the remaining arguments back through the parser, with positional names made into flags (ie `--command ls --args -l`).  I've my new pull request reworks the feature from this one into `unknown-options-as-args`, which will cover the behavior from this pull request, plus the behavior of allowing unknown options to be used as values for known options where `nargs > 0` or the type is `array`.

This new change gets my script working as I would expect it to.

PR #182, opened 2019-06-08
Coerce function returns option of type "number" as "string"

### LINKED ISSUE #176: Coerce function returns option of type "number" as "string"
```js
const parse = require('yargs-parser');

args = parse('--slow 5 --medium 10 --fast 15', {
    number: ['slow', 'medium', 'fast'],
    coerce: {
        'slow': val => val
    },
    configuration: {
        'parse-numbers': true // default anyway
    }
}
);
console.log(args);      // { _: [], slow: '5', medium: 10, fast: 15 }
```

juergba:
### Description

closes #176 

### Description of Change

- `applyCoercions()`: the return value of the coerce function is now checked by calling `maybeCoerceNumber()`, wether it should be converted to type number.
- `maybeCoerceNumber()`: currently return values of coerce functions are not converted to type number.
I deleted this restriction <== **THIS HAS TO BE CONFIRMED**
- extend one existing test

### DIFF
--- a/index.js
+++ b/index.js
@@ -497,7 +497,7 @@ function parse (args, opts) {
   }
 
   function maybeCoerceNumber (key, value) {
-    if (!checkAllAliases(key, flags.strings) && !checkAllAliases(key, flags.coercions)) {
+    if (!checkAllAliases(key, flags.strings)) {
       const shouldCoerceNumber = isNumber(value) && configuration['parse-numbers'] && (
         Number.isSafeInteger(Math.floor(value))
       )
@@ -604,7 +604,7 @@ function parse (args, opts) {
         coerce = checkAllAliases(key, flags.coercions)
         if (typeof coerce === 'function') {
           try {
-            var value = coerce(argv[key])
+            var value = maybeCoerceNumber(key, coerce(argv[key]))
             ;([].concat(flags.aliases[key] || [], key)).forEach(ali => {
               applied[ali] = argv[ali] = value
             })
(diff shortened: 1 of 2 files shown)


juergba:
> [...] because we automatically parse numbers.

It depends on the option type wether we parse or not. The current behavior (without coerce functions) is:
- option is `number`: we parse numbers, configuration `parse-numbers` doesn't matter
- option is `string`: we don't parse numbers, configuration `parse-numbers` doesn't matter
- option is `undefined`: behavior depends on configuration `parse-numbers`

So if users define their option as `string`, they will get  string "55".
It's clearly a bug when the coerce function returns a string into an option of type number.

The key question finally is whether the option type has precedence over the return type of the coerce function.

bcoe:
> The key question finally is whether the option type has precedence over the return type of the coerce function.

Can you explain the bug to me that this solves for you in `mocha`, was coerce returning a number but it gets turned into a string?

juergba:
```js
const parse = require('yargs-parser');

const args = parse('--retries 2 --retries 3', {
    number: ['retries'],
    coerce: {
        'retries': v => Array.isArray(v) ? v.pop() : v
        // 'retries': v => Number(Array.isArray(v) ? v.pop() : v)
    },
    configuration: {
        'parse-numbers': true,               // default
        'duplicate-arguments-array': true    // default
    }
}
);

console.log(args);   // { _: [], retries: '3' }
```
In short: a `number[]` is passed to `coerce` which returns a `string` to a `number` option.
It also returns a `string` when just a `number` is passed to `coerce`.
The alternative with `Number(...)` correctly returns a `number`.

This is not a critical bug for Mocha, so don't wait with cutting your new release.

PR #185, opened 2019-06-25
Fix: boolean[] -  eatArray() with boolean typed arrays and user defaults

juergba:
### Description

```js 
    array: ['file'],
    boolean: ['file'],
    configuration: {
    }
})      // { _: [], file: [ [], [], [] ] }
```
The result is incorrect, it should be: **{ _: [], file: [ [true], [true] ] }**

### Description of Change

- `eatArray()` is called even for flags without arguments. This enables to set:
  - `boolean` with short syntax ('--bool --bool') which carries its implicit value `true`
  - default values given by `opts.default` for all type of options (except `boolean`)
- the fallback default value for all type of options is an empty array `[]`. 
- `eatArray()`: remove the `multipleArrayFlag` logic. The "duplicate array" pattern is already handled within `setKey()` depending on the `flatten-duplicate-arrays` configuration option. `setKey()` remains unchanged.

### DIFF
--- a/index.js
+++ b/index.js
@@ -156,7 +156,7 @@ function parse (args, opts) {
         args.splice(i + 1, 0, m[2])
         i = eatNargs(i, m[1], args)
       // arrays format = '--f=a b c'
-      } else if (checkAllAliases(m[1], flags.arrays) && args.length > i + 1) {
+      } else if (checkAllAliases(m[1], flags.arrays)) {
         args.splice(i + 1, 0, m[2])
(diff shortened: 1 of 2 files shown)


bcoe:
@juergba this is looking like a reasonable start to me, looks like there's an outstanding `TODO`?

bcoe:
@juergba :wave: just following up, would love to see this over the finish line.

bcoe:
@juergba thank you, this looks great.

PR #190, opened 2019-07-23
Added `config-dot-notation` for config files.

f0zi:
In reply to https://github.com/yargs/yargs/issues/1305.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,33 @@ node example.js --foo.bar
 { _: [], "foo.bar": true }
 ```
 
+_Note:_ This setting does not apply to config files (any more). See `config-dot-notation`.
+
+### config-dot-notation
+
+* default: `false`
+* key: `config-dot-notation`
+
+Should keys in the config file that contain `.` split up in objects?
+
+example config.json:
+```json
+{
+  "foo.bar": true
+}
+```
+_if enabled:_
+```sh
+node example.js --config config.json
+{ _: [], foo: { bar: true } }
+```
+
+_if disabled:_
+```sh
+node example.js --config config.json
+{ _: [], "foo.bar": true }
+```
+
 ### parse numbers
 
 * default: `true`
--- a/README.md
+++ b/README.md
@@ -177,14 +177,14 @@ node example.js --foo.bar
 { _: [], "foo.bar": true }
 ```
 
-_Note:_ This setting does not apply to config files (any more). See `config-dot-notation`.
+_Note: as of yargs-parser@14.0.0, `dot-notation` has been split into `dot-notation` and `config-dot-notation`_
 
 ### config-dot-notation
 
 * default: `false`
 * key: `config-dot-notation`
 
-Should keys in the config file that contain `.` split up in objects?
+Should keys in the config file that contain `.` split into objects?
 
 example config.json:
 ```json
--- a/index.js
+++ b/index.js
@@ -16,6 +16,7 @@ function parse (args, opts) {
     'short-option-groups': true,
     'camel-case-expansion': true,
     'dot-notation': true,
+    'config-dot-notation': false,
     'parse-numbers': true,
     'boolean-negation': true,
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -177,6 +177,33 @@ node example.js --foo.bar
 { _: [], "foo.bar": true }
 ```
 
+_Note:_ This setting does not apply to config files (any more). See `config-dot-notation`.
```
bcoe: Perhaps:

`_Note: as of yargs-parser@14.0.0, `dot-notation` has been split into `dot-notation` and `config-dot-notation`_`

README.md
```diff
@@ -177,6 +177,33 @@ node example.js --foo.bar
...
+
+### config-dot-notation
+
+* default: `false`
+* key: `config-dot-notation`
+
+Should keys in the config file that contain `.` split up in objects?
```
bcoe: nit:

> Should keys in the config file that contain `.` be split into objects.

test/yargs-parser.js
```diff
@@ -567,6 +567,7 @@ describe('yargs-parser', function () {
     it('should load nested options from config file', function () {
       var jsonPath = path.resolve(__dirname, './fixtures/nested_config.json')
       var argv = parser(['--settings', jsonPath, '--nested.foo', 'bar'], {
+        configuration: { 'config-dot-notation': true },
```
bcoe: I don't think this logic should need to change because we don't have keys of the form "foo.bar"?

test/yargs-parser.js
```diff
@@ -718,6 +719,7 @@ describe('yargs-parser', function () {
 
     it('should load nested options from config object', function () {
       var argv = parser(['--nested.foo', 'bar'], {
+        configuration: { 'config-dot-notation': true },
         configObjects: [{
           a: 'a',
           nested: {
```
bcoe: It would be good to add two tests, demonstrating "foo.bar" with `config-dot-notation` enabled and disabled.

README.md
f0zi: Ok, I was expecting this :)

README.md
f0zi: NP

test/yargs-parser.js
f0zi: Yeah I was thinking of adding a test but I saw "does not append nested-object keys from config to top-level key" seems to be testing exactly that.

Except, now that I look at it again, the only thing I would change is line 2100 change `'dot-notation': false` to `'config-dot-notation': false` to make it explicit.

Or would you rather like me to add another test case for this?

test/yargs-parser.js
f0zi: Yes, I see what you mean, but it fails without it. Ok, I'll have to look into this one more, tomorrow or Friday.

index.js
```diff
@@ -545,7 +546,7 @@ function parse (args, opts) {
       // if the value is an inner object and we have dot-notation
       // enabled, treat inner objects in config the same as
       // heavily nested dot notations (foo.bar.apple).
-      if (typeof value === 'object' && value !== null && !Array.isArray(value) && configuration['dot-notation']) {
+      if (typeof value === 'object' && value !== null && !Array.isArray(value) && configuration['config-dot-notation']) {
```
bcoe: This logic seems wrong to me, which I believe is why you've had to explicitly turn on `config-dot-notation` to get tests passing.

What I believe we want, when `config-dot-notation` is `false` is to still run `setConfigObject` (which takes care of merging the config file with our arguments.

It's a specific edge-case you want to turn off, which is how someone handles a config file that looks like this:

```
{
  "foo.com": {
     "hello": "world"
  }
}
```

bcoe:
just wanted to bump this PR back up, still something you'd like to see over the finish line @f0zi?

bcoe:
@f0zi closing for now, but please feel free to come back and dust this off when you have time.

PR #198, opened 2019-08-14
Update index.js for option noProcessValue

craigphicks:
Sometimes a command line value not processed by `processValue` is required.
E.g., cannot distinguish between `value` and `"value"` but they could both be valid and carry different meanings for the application.
Initially hoped to modify yargs-parser and yargs so that both values (processed and unprocessed) could be recovered, but the singleton design and interface increase the workload by a large factor compared to this simple edit.

Top level usage example:
```
const argv = yargs
    .command('pp', 'perform reversible pre proc on file', {
        blah blah blah
    })
    .help()
    .alias('help', 'h')
    .parserConfiguration({noProcessValue:true})
    .argv
```
Then the values in argv are raw, e.g., they are as found in process.argv and don't have some quotes elided.

It would be nice to have both raw and processed, either in one call or sequential calls, but attempts to realize that ran into dead ends.  But that's another story.

### DIFF
--- a/index.js
+++ b/index.js
@@ -408,10 +408,10 @@ function parse (args, opts) {
       addNewAlias(key, alias)
     }
 
-    //var value = processValue(key, val)
+    // allow bypassing of processValue with configuration property
     var value = val
-    if (configuration.noProcessValue===undefined || !configuration.noProcessValue){
-      value =processValue(key, val)    
+    if (configuration['no-process-value'] === undefined || !configuration['no-process-value']) {
+      value = processValue(key, val)
     }
 
     var splitKey = key.split('.')
--- a/index.js
+++ b/index.js
@@ -410,9 +410,9 @@ function parse (args, opts) {
 
     //var value = processValue(key, val)
     var value = val
-    if (configuration.noProcessValue===undefined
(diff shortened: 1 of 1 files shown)


bcoe:
@craigphicks curious what your use case is, most shells will drop `"`, so it's not actually possible to pass (to yargs at least) the values `"value"` and `value`.

craigphicks:
With escapes, e.g.,  `"\"value\""`, quotations can be passed.  Currently that requires `"\"\"value\"\""`.  I think what you are saying is that _if an escape is required at all, it's unusable, so the number of escapes required is irrelevant_.

bcoe:
@craigphicks I'm trying to be fairly conservative about adding new configuration settings, and `processValue` does a variety of processing, not just stripping quotation marks.

I'm wondering if there's a specific bug you're seeing in an application that we could perhaps fix in a more targeted way, vs., turning off preprocessing of values entirely.

craigphicks:
I understand your wish to be conservative about new configuration settings.  I'll take care of it on my end.  Thanks for your explanation and thanks for your dedicated hard work.

PR #202, opened 2019-09-25
feat!: rework `collect-unknown-options` into `unknown-options-as-args`, providing more comprehensive functionality

henderea:
@bcoe: since I don't seem to be able to re-open #181, I'm creating a new pull request.  It turns out that the issue I was having with the command parsing after the previous pull request was that it uses an array arg, rather than the `_` array.  I've updated my feature from `collect-unknown-options` to `unknown-options-as-args` so that it can also cover this case.

I've tested this in the script I plan to use it in, and it works as I would expect now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,26 +386,26 @@ node example.js --test-field 1
 { _: [], testField: 1 }
 ```
 
-### collect unknown options
+### unknown options as args
 
 * default: `false`
-* key: `collect-unknown-options`
+* key: `unknown-options-as-args`
 
-Should unknown options be collected into `_`?  An unknown option is one that is not
+Should unknown options be treated like regular arguments?  An unknown option is one that is not
 configured in `opts`.
 
 _If disabled_
 
 ```sh
-node example.js --unknown-option --known-option 2
-{ _: [], unknownOption: true, knownOption: 2 }
+node example.js --unknown-option --known-option 2 --string-option --unknown-option2
+{ _: [], unknownOption: true, knownOption: 2, stringOption: '', unknownOption2: true }
 ```
 
 _If enabled_
 
 ```sh
-node example.js --unknown-option --known-option 2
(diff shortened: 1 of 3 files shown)


bcoe:
@henderea thank you for digging into this, I've been dragging my feet for too long on the next release of yargs, because I was hoping to loop back around to try to dig into this issue.

I've been the victim of my of my libraries' own success these days, and have been spread a bit thin on code review I'm afraid 😦... thanks for your hard work.

@mleguen, could I perhaps loop you in to help review this.

mleguen:
@bcoe my pleasure, but I am afraid I will not be available for a review before wednesday (my weekly OSS day at work).

mleguen:
Wow! Just finished reading the extensive discussion in yargs/yargs#1243, #181 and #202 :-)...
Time to start reviewing!

henderea:
@mleguen On the topic of the forEach stuff, I was largely following the pattern of the existing code.  The comment I put about Node 6 compatibility on the `Object.values` stuff was also copied from the existing code.

I'll make the requested refactoring to reduce the `configuration['unknown-options-as-args'] && isUnknownOption(arg)` into `isUnknownOptionAsArg(arg)` soon.

mleguen:
@henderea OK, I can see now the source of this complicated use of Array.foreach in place of Array.some is checkAllAliases.

I will submit a PR to improve this.

henderea:
@mleguen: I've committed the suggested refactoring of `isUnknownOptionAsArg(arg)`, and also have converted my 2 usages of `forEach` with a boolean to `some`.  I left `checkAllAliases` alone, though, but I can include that in this PR if that would make it easier for everyone.

henderea:
According to https://node.green/, it seems that support for `array.some` goes back just as far as `array.forEach`, so I think it should be fine on v6.

PR #204, opened 2019-10-02
refactoring: remove Array.forEach used in place of Array.some/find

mleguen:
Follow up of https://github.com/yargs/yargs-parser/pull/202#issuecomment-537495395

Replacing Array.forEach in checkAllAliases was not that straightforward, as this function was not really used for what its comment said it was, so I had to use Array.find instead of Array.some.

So I fixed the comment too.


bcoe:
@mleguen mind rebasing? this looks like a smart refactor.

mleguen:
Sure. Added to next wednesday's TODO.

PR #206, opened 2019-10-14
Fix: array option with array default value

### LINKED ISSUE #205: "empty" array values erroring starting with v14.0.0
I checked the changelog, but I couldn't find a match for this behaviour. I have a script on our CI server that evaluates to `--tag --tag smoke` sometimes based on the existence of a env var or not. Essentially `--tag $TAG1 --tag $TAG2`.

This worked as expected in v13.1.1, but started failing in v14.0.0 and v15.0.0 with:

```
Invalid values:
  Argument: tag, Given: , Choices: ...
```

juergba:
closes #205 

Default values for array typed options can be set now as arrays or as primitive values.

### DIFF
--- a/index.js
+++ b/index.js
@@ -388,7 +388,8 @@ function parse (args, opts) {
       // for keys without value ==> argsToSet remains an empty []
       // set user default value, if available
       if (defaults.hasOwnProperty(key)) {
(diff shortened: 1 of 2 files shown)


juergba:
@kellyselden could you give your script a try with this patch, please?

PR #211, opened 2019-10-28
feature: defaulted functionality

### LINKED ISSUE #203: feat: add back defaulted functionality
This was broken, so we removed [here](https://github.com/yargs/yargs-parser/pull/184). We would like to add this functionality back for upstream consumers like Istanbul.

CC: @coreyfarrell

juergba:
closes #203 

`Parse.detailed()` returns an additional object `defaulted` which contains any new argument created by `opts.default`. This argument has not been specified by the user.

` defaulted`: `{ foo: true }`

` defaulted` does not include:
- defaulted aliases
- an option specified by the user on CLI/config/configObjects but without value

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,10 +99,14 @@ yargs engine.
 * `argv`: an object representing the parsed value of `args`
   * `key/value`: key value pairs for each argument and their aliases.
   * `_`: an array representing the positional arguments.
+  * [optional] `--`:  an array with arguments after the end-of-options flag `--`.
 * `error`: populated with an error object if an exception occurred during parsing.
 * `aliases`: the inferred list of aliases built by combining lists in `opts.alias`.
-* `newAliases`: any new aliases added via camel-case expansion.
-* `configuration`: the configuration loaded from the `yargs` stanza in package.json.
+* `newAliases`: any new aliases added via camel-case expansion:
+  * `boolean`: `{ fooBar: true }`
+* `defaulted`: any new argument created by `opts.default`, no aliases included.
+  * `boolean`: `{ foo: true }`
+* `configuration`: given by default settings and `opts.configuration`.
 
 <a name="configuration"></a>
 
--- a/index.js
+++ b/index.js
@@ -35,6 +35,7 @@ function parse (args, opts) {
   var notFlagsOption = configuration['populate--']
   var notFlagsArgv = notFlagsOption ? '--' : '_'
   var newAliases = {}
+  var defaulted = {}
   // allow a i18n handler to be passed in, default to a fake one (util.format).
   var __ = opts.__ || util.format
   var error = null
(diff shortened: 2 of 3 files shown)


bcoe:
@juergba this feature is that simple? wow, I feel a bit silly for the complicated approach we'd left in the codebase for so long :laughing: :cry: 

@coreyfarrell mind taking a look? I believe you were a potential consumer of this functionality.

juergba:
This draft PR is not ready for review, yet. Please note:
- defaulted aliases are not logged
- an option specified on CLI/config/configObjects but without value, is not logged

coreyfarrell:
Quick glance this looks good.  Please comment when you mark this ready for review to make sure github sends notifications.

juergba:
> Please comment when you mark this ready for review to make sure github sends notifications.

Ok, ready for review, thanks.

README.md
```diff
@@ -99,10 +99,14 @@ yargs engine.
...
   * `key/value`: key value pairs for each argument and their aliases.
   * `_`: an array representing the positional arguments.
+  * [optional] `--`:  an array with arguments after the end-of-options flag `--`.
 * `error`: populated with an error object if an exception occurred during parsing.
 * `aliases`: the inferred list of aliases built by combining lists in `opts.alias`.
-* `newAliases`: any new aliases added via camel-case expansion.
-* `configuration`: the configuration loaded from the `yargs` stanza in package.json.
+* `newAliases`: any new aliases added via camel-case expansion:
+  * `boolean`: `{ fooBar: true }`
+* `defaulted`: any new argument created by `opts.default`, no aliases included.
```
bcoe: @juergba what's your reasoning for not including aliases? @coreyfarrell do you think you can work with this limitation?

juergba: `Parse.detailed()` also returns `aliases` which includes all aliases, so the information would be redundant.
It's no big effort to add the aliases to `defaulted`, I just need a decision.

coreyfarrell: For my use it would be fine if `defaulted` did not list aliases, as long as setting an alias causes the original key to be excluded from defaulted.  Would you mind adding a test for this?
```js
    it('setting an alias removes associated key from defaulted', function () {
      var parsed = parser.detailed('--foo abc', {
        default: { kaa: 'abc' },
        alias: { foo: 'kaa' }
      })
      parsed.argv.should.deep.equal({ '_': [], kaa: 'abc', foo: 'abc' })
      parsed.defaulted.should.deep.equal({})
    })
```

Sorry I don't have a chance to actually run this test right now but in theory it should work.

juergba: Done, I added your test.

juergba:
> I think what we might do is introduce the feature `parseDetailed` to yargs, which would have the same output as a detailed parse in `yargs-parser`, [...]

`Yargs.parsed`:
if I understand the Yargs [docu](https://github.com/yargs/yargs/blob/master/docs/api.md#parsed) correctly, then `parsed` should include `defaulted`. I haven't tested it though.

bcoe:
@juergba mind taking a look at https://github.com/yargs/yargs/pull/1472.

PR #221, opened 2019-11-12
Feature: pre-parse string of arguments to argv

### LINKED ISSUE #180: Quotes not consumed
Version: 13.1.0

See also #93 and #138 which both apparently conflict in their expectations of what should happen to quotes.

When I run `parse('a "b c"')` I expect the quotes to be consumed, as happens with most proper shells, but this is not what happens:

```
> parse('a "b c"')
{ _: [ 'a', '"b c"' ] }
```

This expectation appears to be different for different people and I'm not sure if this actually is a bug, but it's not at all what I expected it happen.  Is it possible that we could get an option for how quotes are to be treated?

Here's what I expect to happen based on the behavior of sane shells:

* `a "b c"` becomes `['a', 'b c']`
* `a b\ c` becomes `['a', 'b c']`
* `a '"b c"'` becomes `['a', '"b c"']`
* `a \"b c\"` becomes `['a', '"b', 'c"']`

yargs-parser does something different in all cases. The last one in particular is most surprising.

```
> parse('a "b c"')
{ _: [ 'a', '"b c"' ] }
> parse('a b\\ c')
{ _: [ 'a', 'b\\', 'c' ] }
> parse('a \'"b c"\'')
{ _: [ 'a', '\'"b c"\'' ] }
> parse('a \\"b c\\"')
{ _: [ 'a', '\\"b c\\"' ] }
```

There doesn't seem to be any way to get a literal space into an argument without quotes. If yargs leaves the quotes in, then suddenly it's the responsibility of my application to understand that quotes might be present and strip them. This means **my application is doing the work of the parser**, and that feels very, very wrong -- otherwise, why am I using yargs?

juergba:
### Description

closes #180 

Pre-parse a string of arguments into `argv`, as Node launched in Bash builds `process.argv`.slice(2).
Incl. the handling of quoting and escaped metacharacters.

We add a new method `toArgv()`to `Parser`.

```js
'use strict';
const parser = require('yargs-parser');

let argv = parser.toArgv(String.raw`--foo "bar hello" "testing \" this"`);
console.log(argv);       // [ '--foo', 'bar hello', 'testing " this' ]

let args = parser(argv, {
        array: ['foo']
});
console.log(args);       // { _: [], foo: [ 'bar hello', 'testing " this' ] }
```

Running Node on Bash and calling `parser.toArgv` should result in an identical `argv`: 
```
$ node test.js --foo "bar hello" "testing \" this"
// process.argv.slice(2) === [ '--foo', 'bar hello', 'testing " this' ]
```
```
// let res = parser.toArgv(`--foo "bar hello" "testing \\" this"`)  // ok, works
let res = parser.toArgv(String.raw`--foo "bar hello" "testing \" this"`)
// res === [ '--foo', 'bar hello', 'testing " this' ]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,18 @@ console.log(argv)
 { _: [], foo: 99, bar: 33 }
 ```
 
+Pre-parse a complex string with escaped metacharacters before passing to `yargs-parser`:
+```js
+const parse = require('yargs-parser')
+let argv = parse.toArgv(String.raw`--foo "hello world" "testing \" \"quotes\""`)
+let args = parse(argv, { array: ['foo'] })
+console.log(args)
+```
+
+```sh
+{ _: [], foo: [ 'hello world', 'testing " "quotes"' ] }
+```
+
 Convert an array of mixed types before passing to `yargs-parser`:
 
 ```js
@@ -124,7 +136,7 @@ var parsed = parser(['--no-dice'], {
 })
 ```
 
-### short option groups
+#### short option groups
 
 * default: `true`.
 * key: `short-option-groups`.
@@ -143,7 +155,7 @@ node example.js -abc
 { _: [], abc: true }
 ```
 
-### camel-case expansion
+#### camel-case expansion
 
 * default: `true`.
 * key: `camel-case-expansion`.
@@ -162,7 +174,7 @@ node example.js --foo-bar
 { _: [], 'foo-bar': true }
 ```
 
-### dot-notation
+#### dot-notation
 
 * default: `true`
 * key: `dot-notation`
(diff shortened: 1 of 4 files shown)


bcoe:
:tada: thanks for taking this on @juergba, @cdhowie do you think this would work for your use-case.

bcoe:
@juergba coming back to this, I would personally rather we just document [arrgv](https://github.com/astur/arrgv) in the README, with a clear example of how one would use this to parse a complex argument string, rather than introducing an additional dependency to the chain.

At the end of the day, someone just invokes `aargv`, might as well just point folks towards that dependency?

juergba:
@bcoe yes, I can live with your decision.

I leave the README update to someone else though, I don't have time for this right now.
So I'm closing this PR.

bcoe:
@juergba 👌 appreciate the work you put into this.

PR #222, opened 2019-11-17
Trailing and leading quotes

### LINKED ISSUE #201: Trailing and leading quote characters are removed
Version 13.1.1

### Description

Trailing and leading quote characters are removed from the values. If I want to pass a `'string'` as a value I need to triple-quote it as `"''string''"` or `\'''string\'''` when I'm specifying that in the shell.

### Expected

```javascript
> require('yargs-parser')([ '--arg', '"option"' ])
{ _: [], arg: '"option"' }
```

### Actual

```javascript
> require('yargs-parser')([ '--arg', '"option"' ])
{ _: [], arg: 'option' }
```

This probably relates to #138, #180. Definitely relates to #145 and #153.
I would suggest introducing a new `keepQuotes`/`trimQuotes` option to support the all edge cases.

juergba:
### Description

closes #201 

Currently trailing and leading quotes are removed, this applies to `string` and also to `string[]` input.
```js
> require('yargs-parser')([ '--arg', '"option"' ])
{ _: [], arg: 'option' }
```
One (main?) reason for this: `--arg "--option"`
if user needs `--option` to be handled as value (no option), he wraps the value into quotes which are removed by yargs-parser later on ==> `{ _: [], arg: '--option' }`
Problem: each and every tl-quotes are removed.

### Description of change

This is a delicate topic, so the following is a proposition.

- we distinguish between `string` input and `string[]` input.
  - `string`: there are good reasons to remove tl-quotes, as does eg Bash.
  - `string[]`: we never remove any tl-quotes.
- therefore we move the quote handling from "index.js" to "tokenize-arg-string.js".
- we introduce the escaped dash '\\\\-': 
`[ '--arg', '\\--option' ]` ==> `{ _: [], arg: '--option' }`

BREAKING CHANGE

### DIFF
--- a/test/yargs-parser.js
+++ b/test/yargs-parser.js
@@ -3270,8 +3270,8 @@ describe('yargs-parser', function () {
       args.foo.should.equal('hello world')
       args.bar.should.equal('goodnight\'moon')
       const args2 = parser(['--foo', '"hello world"', '--bar="goodnight\'moon"'])
-      args2.foo.should.equal('hello world')
-      args2.bar.should.equal('goodnight\'moon')
+      args2.foo.should.equal('"hello world"')
+      args2.bar.should.equal('"goodnight\'moon"')
     })
 
     it('handles single quoted strings', function () {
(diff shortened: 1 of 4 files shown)


test/yargs-parser.js
```diff
@@ -3270,23 +3270,26 @@ describe('yargs-parser', function () {
...
+      args2.foo.should.equal("'hello world'")
+      args2.bar.should.equal("'goodnight\"moon'")
     })
 
     it('handles strings with dashes', function () {
-      const args = parser('--foo "-hello world" --bar="--goodnight moon"')
+      const args = parser('--foo "\\-hello world" --bar="--goodnight moon"')
       args.foo.should.equal('-hello world')
       args.bar.should.equal('--goodnight moon')
+      const args2 = parser(['--foo', '\\-hello world', '--bar=--goodnight moon'])
```
bcoe: I don't love the `\\` syntax, is there prior art this is based on?

PR #236, opened 2020-02-09
feat!: boolean arguments will not be collected into an implicit array

bcoe:
This reverts behavior to match yargs@14. If an argument is configured as a `bool` we will not coerce it into an array.

@SimenB could I get you to try out this branch if you have a moment.

Refs: https://github.com/facebook/jest/issues/9517
BREAKING CHANGE: this reverts parsing behavior of booleans to yargs@14


SimenB:
Yeah, this works as expected. Thanks!

PR #271, opened 2020-05-21
fix: back-porting prototype fixes for *really* old version

phated:
GitHub won't let me create a draft against another tag so this diff is going to be really weird.

I'm guessing the best way to work on this would be to branch `v5.0.0` tag into a branch and then PR against that. Once it is merged and released, the branch can be cleaned up since the tag will point to the new version.

Drafting while WIP.


phated:
@bcoe whatcha think?

phated:
I'll squash this to get CI to run now that I've switched the base branch.

Issue #265, opened 2020-04-02
Switch to typescript


mleguen:
Needed for yargs/yargs#1586

QmarkC:
I would like to help with this. I reviewed yargs/yargs#1586 and the related pull requests. Based on that, I propose the first step be the initial TypeScript configuration and converting `lib/tokenize-arg-string.js` to TypeScript. This is similar to the [first step PR](https://github.com/yargs/yargs/pull/1587) for yargs/yargs#1586.

I created a draft PR (#272) with this proposed first step to make feedback/discussion a little easier for any of those changes.

Beyond the first step, is there any guidance on how to proceed?
To avoid big-bang merge requests I imagine you would want to convert `index.js` in parts to `lib/*.ts` files.
I created the following list with all the functions in `index.js` and took a first pass at grouping them together by topic/concern. Each group could be their own lib module or some you might prefer to be their own module (e.g. `is-number.ts` vs a `utils.ts` that exports `isNumber`) since yargs/yargs has a `is-promise.ts` module.

- Parser
   - Parser
   - Parser.detailed
- configuration
    - defaultConfiguration
    - checkConfiguration
    - setConfig
    - setConfigObject
    - setConfigObjects
- parse
    - parse
    - applyDefaultsAndAliases
- parse-argv
    - argv variable assignment in parse
    - isUnknownOptionAsArg
    - isUnknownOption
    - setArg
    - setPlaceholderKeys
    - applyCoercions
    - applyEnvVars
    - eatArray
    - eatNargs
    - processValue
    - maybeCoerceNumber
    - defaultValue
    - guessType
    - defaultForType
- parse-aliases
    - addNewAlias
    - combineAliases
    - checkAllAliases
    - extendAliases
- parse-flags
    - hasAnyFlag
    - hasFlagsMatching
    - hasAllShortFlags
- utils
    - isNumber
    - isUndefined
    - sanitizeKey
    - hasKey
    - setKey
    - increment

bcoe:
If @mleguen is open to the help with the conversion, to me it seems like it would be great to have another person pitching in.

mleguen:
Thanks @QmarkC for the help!

The approach I took (and try to hold) for yargs, is to focus only on switching to TS, and not to refactor at the same time, as it is already difficult enough to review as is.

So I would recommend keeping at the moment a single `lib/index.ts` module, and refactor it later into several modules in separate issues/PR.

QmarkC:
[...]

PR #272, opened 2020-05-21
chore(ts): init typescript conf + tsify tokenize-arg-string

QmarkC:
First step of #265, "Switch to typescript".

- Add initial TypeScript configuration
  Modeled after TypeScript configuration used in [yargs/yargs](https://github.com/yargs/yargs).

  Matched semantic versions used in [yargs package.json](https://github.com/yargs/yargs/blob/master/package.json) for added devDependencies.

  - Add/Update configs related to `build` output folder.
  - Move coverage thresholds from npm script to `.nycrc` config

- Add lint configuration for `.ts` files.
  Modeled after yargs setup.
  - Switch from `standard` to `standardx`
  - Add `.editorconfig`
  - Add ESLint config

- Convert `lib/tokenize-arg-string.js` and `test/tokenize-arg-string.js` to TypeScript
- Update `index.js` to use `tokenizeArgString` from build


.eslintrc
```diff
@@ -0,0 +1,21 @@
+{
+  "overrides": [
+    {
+      "files": "*.ts",
+      "parser": "@typescript-eslint/parser",
+      "rules": {
+        "no-unused-vars": "off",
+        "no-useless-constructor": "off",
+        "@typescript-eslint/no-unused-vars": "error",
```
bcoe, line 9: I'd like to move towards using [gts](https://www.npmjs.com/package/gts) rather than a custom `.eslintrc`, but this will be a breaking change.

@mleguen realistically, do you think we'll be able to convert `yargs-parser` and `yargs/yargs.js` without calling it breaking?

I'm wondering if we should release a new version of yargs, with what's been converted so far, and plan on this final work being a major bump.

lib/tokenize-arg-string.ts
```diff
@@ -1,9 +1,15 @@
 // take an un-split argv string and tokenize it.
-module.exports = function (argString) {
+export function tokenizeArgString (argString: string | any[]): string[] {
   if (Array.isArray(argString)) {
     return argString.map(e => typeof e !== 'string' ? e + '' : e)
   }
 
+  // For complied output, typescript prevents this type guard branch
+  /* c8 ignore next 3 */
+  if (typeof argString !== 'string') {
```
bcoe, line 9: I think rather than this throw, you could just do this:

```
cons argString = maybeArgString as string[]
```

lib/tokenize-arg-string.ts
```diff
@@ -1,9 +1,15 @@
 // take an un-split argv string and tokenize it.
-module.exports = function (argString) {
+export function tokenizeArgString (argString: string | any[]): string[] {
```
bcoe, line 2: perhaps call this `maybeArgString`.

package.json
```diff
@@ -4,10 +4,14 @@
   "description": "the mighty option parser used by yargs",
   "main": "index.js",
   "scripts": {
-    "fix": "standard --fix",
-    "test": "c8 --reporter=text --reporter=html  mocha test/*.js",
-    "posttest": "standard",
-    "coverage": "c8 report --check-coverage check-coverage --lines=100 --branches=97 --statements=100"
+    "fix": "standardx --fix && standardx --fix **/*.ts",
```
bcoe, line 7: depending on @mleguen's feedback, I might be tempted to use `gts` here.

test/tokenize-arg-string.ts
```diff
@@ -1,58 +1,58 @@
 /* global describe, it */
+import { expect, should } from 'chai'
+import { tokenizeArgString } from '../lib/tokenize-arg-string'
 
-const tokenizeArgString = require('../lib/tokenize-arg-string')
+should()
```
bcoe, line 5: let's just switch to using `expect`, if it's not too much of a hassle, rather than doing back flips to address linting.

QmarkC: > let's just switch to using `expect`, if it's not too much of a hassle, rather than doing back flips to address linting.

Done.

lib/tokenize-arg-string.ts
QmarkC: I'll update this to use a type assertion based on the suggestion. Below I just wanted to share my thought process when working on the original PR.

Some kind of type protection needs to be added for when the tokenizeArgString argument is not a string or an array for TypeScript to compile since `trim` and `charAt` does not exist on an array type (e.g. `any[]`).

When I was considering this scenario I came up with the following options:
1. Add a type assertion. E.g. `argString = (argString as string).trim()`
  This is similar to what @bcoe suggested but I think it should be as a `string` not a string array `string[]` since the `Array.isArray(argString)` branch would handle the array case. This results in the same output and behavior as current, with a TypeError getting thrown if the argument is not a string or array.
1. Throw an Error with an invalid argument type of message
    - A TypeError would match current nodejs behavior.
    - Error would be better supported for non-nodejs env.
    - Add a custom error class, same or similar to [YError](https://github.com/yargs/yargs/blob/master/lib/yerror.ts) in yargs.
1. Use same logic as when an array element is not type of string.
    - This adds some consistently to how non-strings are handled but I'm not sure if that would be considered a breaking change. This would also require changing the argument type.
    - One could argue that it is not a breaking change since tokenizeArgString is not exposed from the main module and due to the `slice` call before the argument is passed to tokenizeArgString it is not really possible to pass a non-string or non-array as an argument. I guess technically Blob or any custom object w/ a slice method might get passed too.
1. Return Error rather than throwing one but this would require other changes such as:
    - Move the error variable declaration in index.js above the tokenizeArgString call
    - Add check if args is instance of Error, if true then assign to the error variable
    - Then decide how to handle the use case from there such as: set args to empty array, wrap rest of the required logic in parser with if statement, or return early.
1. Return an empty array.

I opted for option 2, throwing an Error with an invalid argument message, similar to [yargs argument type error](https://github.com/yargs/yargs/blob/master/lib/argsert.ts#L63-L65) since this seemed like the most type safe with the smallest amount of changes. This is also similar to the current behavior but with a more specific error message describing the issue. The disadvantage of this and some of the other options is getting full test coverage since the types prevent non-string or non-array argument from being passed. At least w/o jumping through hoops such as changing types during the test and/or adding more advanced mapped types.

lib/tokenize-arg-string.ts
QmarkC: I was just keeping with the existing argument name. It does keep a nice consistency with the function name though. I really don't have a strong preference either way. 

My thought on if changing the argument name to `maybeArgString` would be to update the first few lines like so:

```js
export function tokenizeArgString (maybeArgString: string | any[]): string[] {
  if (Array.isArray(maybeArgString)) {
    return maybeArgString.map(e => typeof e !== 'string' ? e + '' : e)
  }

  const argString = (maybeArgString as string).trim()
```

Another option would be to drop the type from name. It's just called `args` in parse when calling `tokenizeArgString(args)`. Then to keep the same type of naming consistency could change the function name to `tokenizeArgs`. This would be my personal preference.

Just let me know what y'all want and I'll update as needed.

.eslintrc
QmarkC: There are quite a few style differences when using `gts`. Such as:
- semicolons
- trailing commas
- bracket spacing
- arrow body style
- keyword spacing

There would be over 3,000 problems that would need to be fixed. Most are automatically fixable with the `--fix` option. However this seems like the type of big-bang PR you are wanting to avoid. So you might consider a more gradual transition such as only applying it in the `lib` folder or on a per rule type of change like adding semicolons. Then once all the code has been updated to `gts` doing a cleanup of the lint config to just have `gts` related config.

mleguen: I recommend:
- switching to ts first, keeping eslint at the moment
- then drop nodejs 8 support and switch to gts for linting, in a carefully reviewed big-bang PR. I already did a "dry-run" of this on yargs a few months ago, and it is not that difficult to review. IMHO doing it gradually would bring more cons than pros.

mleguen: > realistically, do you think we'll be able to convert yargs-parser and yargs/yargs.js without calling it breaking?

I'd like too. But I will know for sure if it is possible when I will start working on yargs' index.d.ts

lib/tokenize-arg-string.ts
mleguen: I recommend using `myValue as MyType` only as a last resort, and only to solve cases when you are sure `myValue` has indeed type `MyType`, and it is typescript which does not manage to infer correctly the type.

*This is not the case here.*

If typescript is complaining about a type, and is right about doing so, the most elegant and readable way to address this kind of typing issue, IMHO, is to encapsulate a `if/throw` - or better, an assertion from an assertion lib, as it does not count as an uncovered branch - into a function with the special typescript return value `assert xxx is yyy`.

*This is still not the case here.*

In fact, typescript does not complain and already considers `argString` as a string here, so I would not add anything. Perhaps an input argument check is missing, but adding missing type checks is not our goal in this PR, only switching to typescript, and changing as little code as it is possible doing so for it to be easily reviewed.

lib/tokenize-arg-string.ts
mleguen: I would recommend not to rename functions or arguments, once again to limit code changes.

test/tokenize-arg-string.ts
mleguen: I recommend keeping tests as they are written, some with should other with expect. Once again, we should not mix switching to ts and refactoring (especially refactoring tests).

lib/tokenize-arg-string.ts
QmarkC: @mleguen I agree with using a type assertion as a last resort. TypeScript does consider `argString` a string with the type `string | any[]` but does throw the described error `TS2339` with a read only array `string | ReadonlyArray<any>`, which is why I originally had the type guard.

So what type would you like to use for the `argString` argument?
I chose `string | any[]` since it seemed like the best fit without any possible breaking changes. However I would prefer to avoid using `any` and think a read only array is a better choice over array since the array is not modified.

Currently:

- The yargs-parser README states: `args` should be a string or array of strings. `string | string[]`
- The [types for yargs-parser](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/yargs-parser/index.d.ts#L100) uses `string | string[]` for the Parser argv argument.
- [Parser](https://github.com/yargs/yargs-parser/blob/master/index.js#L1015) calls `.slice()` on the argv argument. Then argv gets [passed to tokenizeArgString](https://github.com/yargs/yargs-parser/blob/master/index.js#L11). That confirms string or array.
- In tokenizeArgString when [argString is an array](https://github.com/yargs/yargs-parser/blob/master/lib/tokenize-arg-string.js#L4), any array elements that are not type of string get converted to a string.
- There is also a [test case](https://github.com/yargs/yargs-parser/blob/master/test/tokenize-arg-string.js#L15) for passing a number as part of an array. That suggests at least `string | Array<string | number>`.
- However really any other type will get `.toString()` applied with `e + ''` as well. (null => 'null', {} => '[object Object]', [] => '', etc...). Which is what lead me to using `string | any[]`.

So that suggests the following options:
- `string | any[]`
- `string | string[]`
    - This seems like would it should be based on the current README and types but is likely a breaking change as it breaks the test case with a number in an array.
- `string | Array<string | number>`
    - Including number as type in the array makes me question if the other yargs option types should also be supported. Yargs has `boolean`, `number`, `string`, `array`, and `count` (which would be a number type). So perhaps `string | Array<string | number | boolean>` or with nested array `string | Array<string | number | boolean | Array<string | number | boolean>>`
- Using `ReadonlyArray` rather than `Array` for any of the above. Which would require a type guard like I originally had in this PR.

test/tokenize-arg-string.ts
QmarkC: @mleguen are you ok with how I originally had the tests? This was mostly the same as they were written. I just added a couple new test cases and grouped the test cases based on the argument type.

lib/tokenize-arg-string.ts
bcoe: we enforce arguments with a throw in most public methods on `yargs` itself, I'm not opposed to this. It would just potentially need to be a breaking change, since it changes the API surface... In practice it might just throw an exception if you pass in other arguments any ways...

But, if we were to add in the throw, I'd suggest we just add a unit test for it rather than skipping with c8.

mleguen: > TypeScript does consider argString a string with the type string | any[] 

Before `if (Array.isArray(argString)) return...`, but not after, as `argString` cannot be an array anymore.

argstring

So there is really no need for this if/throw here.

package.json
mleguen: In a separate PR, perhaps?

test/tokenize-arg-string.ts
mleguen: > I just added a couple new test cases

No problem with this

> and grouped the test cases based on the argument type

Maybe should we separate this refactoring?

.versionrc
```diff
@@ -0,0 +1,9 @@
+{
+  "types": [
+    {"type":"feat","section":"Features"},
+    {"type":"fix","section":"Bug Fixes"},
```
bcoe, line 4: the [tool we're using for releasing](https://github.com/googleapis/release-please) doesn't read this file format ... might actually be a good idea to add support.

But might as well drop this file for now.

bcoe:
@QmarkC this is looking pretty solid to me, happy to take the PR whenever you're ready.

lib/tokenize-arg-string.ts
QmarkC: @mleguen the error is thrown when using ReadonlyArray, but is not thrown when using Array.

.versionrc
QmarkC: @bcoe Done. I didn't see that documented any where so I just assumed it was the same as yargs.

test/tokenize-arg-string.ts
QmarkC: @mleguen updated as recommenced.

QmarkC:
> @QmarkC this is looking pretty solid to me, happy to take the PR whenever you're ready.

I've updated the PR based on the current feedback.
There is still an [open question](https://github.com/yargs/yargs-parser/pull/272#discussion_r431216196) regarding the type to use the `argString` argument. I would like to get confirmation on using `string | any[]` or if this should be more strict such as `string | (string | number)[]`. The same type will need to be used for `Parser`, `Parser.detailed`, and `parse`.

Also should I rebase to merge all the commits for cleaner history? That's what I typically do after getting feedback on a PR.

lib/tokenize-arg-string.ts
QmarkC: I wasn't clear on why that was so I [asked on stack overflow](https://stackoverflow.com/questions/62133217/error-ts2339-whats-the-difference-when-using-array-vs-readonlyarray).
`ReadonlyArray` could be used without the type guard if we rearraged it handle the string case first before the array.

```ts
export function tokenizeArgString (argString: string | ReadonlyArray<any>): string[] {
  if (typeof argString === 'string') {
    argString = argString.trim()

    let i = 0
    let prevC: string | null = null
    let c: string | null = null
    let opening: string | null = null
    const args: string[] = []

    for (let ii = 0; ii < argString.length; ii++) {
      prevC = c
      c = argString.charAt(ii)

      // split on spaces unless we're in quotes.
      if (c === ' ' && !opening) {
        if (!(prevC === ' ')) {
          i++
        }
        continue
      }

      // don't split the string if we're in matching
      // opening or closing single and double quotes.
      if (c === opening) {
        opening = null
      } else if ((c === "'" || c === '"') && !opening) {
        opening = c
      }

      if (!args[i]) args[i] = ''
      args[i] += c
    }

    return args
  }

  return argString.map(e => typeof e !== 'string' ? e + '' : e)
}
```

test/tokenize-arg-string.ts
QmarkC: Created a new issue to switch tests to use expect #275 to capture this feedback.

bcoe:
@QmarkC I would like to stick with the `string | any[]`, as other changes would be breaking (the `"" + arg` is coercing any types into a string any ways). _The current approach looks good to me._

Looks like your rebased already, in general this is optional, as I squash and merge when I land 👍 I sometimes do so any ways on a large feature.

bcoe:
we can land this as soon as @mleguen approves. Thanks for getting this work started @QmarkC.

bcoe:
@QmarkC I've been talking with @mleguen, I think we'll make the final TypeScript release of yargs a major release any ways... so, let's plan that this work on `yargs-parser` will be a major, at which point let's make a PR to drop Node 8, move to Node 10, and move to the latest version of decamelize.

The problem with decamelize@3.x is the `xregexp` dep, which is 700kb.

package.json
```diff
@@ -27,20 +31,38 @@
...
   },
   "files": [
-    "lib",
-    "index.js"
+    "index.js",
+    "build",
+    "lib/**/*.js"
   ],
   "engines": {
     "node": ">=6"
```
mleguen, line 60: If we do want to continue supporting node 6 in `yargs-parser`, then the target needs to be changed in `tsconfig.json` (and in `.eslintrc` too, I think), as node 6 supports only 17% of es2017 (see https://node.green/#ES2017).

mleguen: But as I see we no longer do tests on node 6, I suggest upgrading the engines requirements to `"node": ">=8"` here

QmarkC: https://github.com/yargs/yargs-parser/pull/278

mleguen: OK, but opening #278 does not change the issue here: tsconfig.json and .eslintc are still incoherent with package.json?

bcoe:
@QmarkC master has the latest decamelize and camelcase now 👍 

Resolve conflicts, make sure you've addressed @mleguen's feedback, and we'll merge this?

package.json
QmarkC: @mleguen engines.node in package.json has been updated to `>=10`, ESLint `parserOptions.ecmaVersion` updated to `2018`, and updated tsconfig.json to remove `es2017` so it will use `es2018` from `gts/tsconfig-google.json`.

QmarkC:
> Thanks a lot! Ready to merge once engines requirements in `package.json` is in line with tsc output.

@bcoe @mleguen this should be resolved now with the update to node `>=10`, and `es2018` for ESLint and tsconfig.json.

I also created another branch, [265-tsify-yargs-parser](https://github.com/yargs/yargs-parser/compare/master...QmarkC:265-tsify-yargs-parser), that has this first step, second step to tsify yargs-parser, and cleans up the commits. Let me know if you want to close this PR in favor of that branch.

PR #277, opened 2020-06-03
Revert "fix(deps): update dependency decamelize to v3"

bcoe:
Reverts yargs/yargs-parser#274

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "camelcase": "^5.0.0",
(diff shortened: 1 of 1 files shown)


bcoe:
@QmarkC I landed #274 prematurely, the issue with pulling in a new decamelize is the dependency `xregexp`, which does exist in `3.2.0 -- _I remember now why I was holding off on upgrading decamelize._

`xregexp` is `699.5kb` unpacked, 6x the times of yargs-parser itself.

PR #278, opened 2020-06-04
build!: drops Node 6. begin following Node.js LTS schedule

bcoe:
This skips the EOL Node 8, and moves us to Node 10 as the supported Node.js version for yargs-parser.

_Note: we should complete the TypeScript conversion before we release this major version of the library._

TODO:

- [x] I'd like to add a snippet of code that throws if you're on an older Node.js version.

Refs #272

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [8, 10, 12, 13]
+        node: [10, 12, 14]
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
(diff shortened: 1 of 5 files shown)


QmarkC:
> TODO:
> 
> * [ ]  I'd like to add a snippet of code that throws if you're on an older Node.js version.

Run time or build time e.g. in pretest?
Basic check such as parsing [`process.version`](https://nodejs.org/docs/latest-v10.x/api/process.html#process_process_version) or add a dep like [`check-node-version`](https://www.npmjs.com/package/check-node-version)?

bcoe:
@QmarkC I think it's worth adding something along the lines of:

```
const minNodeVersion = 10;
if (process && process.version) {
  const major = Number(process.version.match(/v([^.]+)/)[1]);
  if (major < minNodeVersion) {
    throw Error(`yargs parser supports a minimum Node.js version of ${minNodeVersion}`)
  }
}
```

PR #295, opened 2020-07-30
feat!: adds support for ESM and Deno

bcoe:
Using @mleguen and @QmarkC's work as a starting point, this refactor adds support for ESM and Deno.

The approach taken is as follows:

1. TypeScript now targets es2015 module format.
2. rollup is used to create a bundled CJS build of the library, including an `index.cjs.d.ts` (which is the binding that will be used for `require` statements).
3. conditional exports are used to target both ESM and CJS.

## Todo:

- [x] finish updating tests for `tokenize-arg-string.ts`.
- [x] document in README how to use deno and ESM.
- [x] investigate publishing to pika (I've added a _module_ field, which I believe should allow us to start showing up on pika).
- [x] add config loading support to Deno.
- [x] make config loading work for ESM module.
- [x] make config loading throw reasonable error in browser.
- [x] get test coverage working for bundle.
- [x] configure release-please to push deno branch on release.

BREAKING CHANGE: `decamelize` and `camelCase` logic has been re-implemented, and may have some logical differences. Conditional exports for CJS vs., ESM may break some environments.


scripts/replace-legacy-export.cjs
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
+'use strict'
+
+const { readFileSync, writeFileSync } = require('fs')
+
+// Cleanup the export statement in CJS typings file generated by rollup:
+const legacyTypings = 'build/index.cjs.d.ts'
+const contents = readFileSync(legacyTypings, 'utf8').replace(
+  'export { yargsParser as default };', 'export = yargsParser;'
```
jkrems, line 9: Is this a bug in rollup's support for `--exports=default`? If so, is there an open issue this could link to track when the workaround may become unnecessary?

bcoe: @jkrems I wanted to keep the behavior of the CJS module identical to what it is today, with the main export being a callable function, i.e., not `exports.default = yargsParser`. However, for ESM/Deno I wanted to `export default yargsParser`... I wasn't sure how to indicate to rollup that I wanted the `export = yargsParser` behavior, while at the same time using the desired syntax in the source files themselves.

Maybe I should open a ticket on the tsc rollup plugin I'm using?

jkrems: Yeah, it seem like if the `module.exports` object is set to the default export, then `export = defaultExport` would be the correct typings to emit. At least from what I understand about TS type definitions. Not sure how that would affect future ESM consumers (that use vscode code completion) in the presence of named exports etc. but that's not necessarily a problem for today..?

scripts/replace-legacy-export.cjs
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
+'use strict'
+
+const { readFileSync, writeFileSync } = require('fs')
+
+// Cleanup the export statement in CJS typings file generated by rollup:
+const legacyTypings = 'build/index.cjs.d.ts'
```
evocateur, line 7: I'm very leery of implicit path locations, silently dependent on "where" one executes the code from.

```suggestion
const legacyTypings = require('path').resolve(__dirname, '../build/index.cjs.d.ts')
```

package.json
```diff
@@ -2,15 +2,25 @@
...
+    "pretest": "rimraf build && tsc -p tsconfig.test.json && npm run build:cjs",
     "test": "c8 --reporter=text --reporter=html mocha test/*.js",
     "posttest": "npm run check",
     "coverage": "c8 report --check-coverage",
-    "check": "standardx ./*.ts && standardx **/*.ts",
-    "compile": "rimraf build && tsc",
+    "precompile": "rimraf build",
+    "compile": "tsc",
+    "postcompile": "npm run build:cjs",
+    "build:cjs": "rollup --exports=default --config=rollup.config.js ./index-node.ts --format cjs --file build/index.cjs",
```
evocateur, line 22: A lot of these arguments could be configured more expressively in `./rollup.config.js`?
```suggestion
    "build:cjs": "rollup -c",
```
```js
export default {
  input: './index-node.ts',
  output: {
    format: 'cjs',
    file: './build/index.cjs',
    exports: 'default',
  },
  plugins: [
    ts({ /* options */ })
  ],
}
```

lib/yargs.ts
```diff
@@ -0,0 +1,1101 @@
...
+  StringFlag,
+  BooleanFlag,
+  NumberFlag,
+  ConfigsFlag,
+  CoercionsFlag,
+  Options,
+  OptionsDefault,
+  YargsParserMixin
+} from './yargs-parser-types.js'
+import type { Dictionary, ValueOf } from './common-types.js'
```
evocateur, line 24: Aren't these files `.ts`? I'm confused by the explicit `.js` extensions.

rollup.config.js
```diff
@@ -0,0 +1,9 @@
+import ts from '@wessberg/rollup-plugin-ts'
+
+export default {
+  plugins: [
```
evocateur, line 4: I think you might be missing [`@rollup/plugin-node-resolve`](https://github.com/rollup/plugins/tree/master/packages/node-resolve), configured with `extensions: ['js', 'ts']`?

lib/yargs.ts
bcoe: this is an oddity I learned during this process, TypeScript doesn't allow you to include `.ts` extensions, they need to instead reference the path that will eventually be the target extension.  So, to target code that works for ESM and Deno, you need to include the `.js` extension, which will then be included in the compiled code.

evocateur: Wow, TIL.

Issue #345, opened 2020-12-09
env() removes capitalization from camel case variable names


pownkel:
In the update to v16, it looks like the behavior of env() changed to convert all env variable names to lowercase. Here's an example to illustrate:

```
process.env.camelCaseArg = 'value';
process.env.lowercasearg = 'value';

const envArgs = yargs().env().argv;

console.log(`lower case: ${envArgs.lowercasearg}`);
console.log(`camel case: ${envArgs.camelCaseArg}`);
console.log(`camel case variable with lower case name: ${envArgs.camelcasearg}`);

```

Running this on v15.4.0 prints
```
lower case: value
camel case: value
camel case variable with lower case name: undefined
```
While in v16.0.0 and later, the same code prints
```
lower case: value
camel case: undefined
camel case variable with lower case name: value

```
Updating to v16 broke some of our code as our camel case env variables were no longer found under their original names. I'm not sure whether the change was intentional, but the [API documentation for .env()](https://github.com/yargs/yargs/blob/master/docs/api.md#envprefix) still shows camel case env variable names in the argv object, which is misleading.

bcoe:
@pitgrap @pownkel thank you for the bug report, I'm pretty sure this will end up being related to my reimplementing our camel case logic in v16 -- if you get a chance to dig into it, patches are very much welcome.

pitgrap:
@bcoe can you give me a code pointer? I tried to find the PR, which introduced this bug, but gave up after 10min search. :-/

bcoe:
@pitgrap I implemented camel-case, and de-camel-case here:

https://github.com/yargs/yargs-parser/blob/master/lib/string-utils.ts#L26

When I ported the library to ESM.

Ivan-Strahovsky:
@bcoe I have fix in my local branch, how can I open PR for to review?

bcoe:
this should be fixed in the latest release.

PR #348, opened 2021-02-02
#345 do not lowercase camel cased string

Ivan-Strahovsky:
PR to fix issue https://github.com/yargs/yargs-parser/issues/345

### DIFF
--- a/lib/string-utils.ts
+++ b/lib/string-utils.ts
@@ -1,4 +1,6 @@
 export function camelCase (str: string): string {
+  // Handle the case where an argument is provided as camel case, e.g., fooBar.
+  // by ensuring that the string isn't already mixed case:
   const isCamelCase = str !== str.toLowerCase() && str !== str.toUpperCase()
 
   if (!isCamelCase) {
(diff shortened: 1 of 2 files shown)


test/string-utils.cjs
```diff
@@ -11,6 +11,12 @@ describe('string-utils', function () {
     it('removes leading hyphens', () => {
       strictEqual(camelCase('-goodnight-moon'), 'goodnightMoon')
     })
+    it('camelCase string stay as is', () => {
```
bcoe, line 14:
```suggestion
    it('camelCase string stays as is', () => {
```

lib/string-utils.ts
```diff
@@ -1,5 +1,10 @@
 export function camelCase (str: string): string {
-  str = str.toLocaleLowerCase()
+  const isCamelCase = str !== str.toLowerCase() && str !== str.toUpperCase()
```
bcoe, line 2:
```suggestion
  // Handle the case where an argument is provided as camel case, e.g., fooBar.
  // by ensuring that the string isn't already mixed case:
  const isCamelCase = str !== str.toLowerCase() && str !== str.toUpperCase()
```

bcoe (changes requested):
Added two suggestions for you, thanks for this patch.

bcoe (approved):

PR #358, opened 2021-03-02
docs: correct typo in readme examples

zypA13510:
On my end:
```console
$ node -e 'console.log(require("yargs-parser")(process.argv.slice(1)))' 99.3
{ _: [ 99.3 ] }
```

Not sure why it is 99 in the [README](https://github.com/yargs/yargs-parser/blame/9cec00a622251292ffb7dce6f78f5353afaa0d4c/README.md#L255), seems to me like a typo.

Also updated all the example snippets to use the correct language code "console" (which is for shell script and their output) instead of "sh" (which is for shell script only). [[rendered preview](https://github.com/zypA13510/yargs-parser/blob/patch-1/README.md)]

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ Should positional keys that look like numbers be treated as such.
 
 ```console
 $ node example.js 99.3
-{ _: [99] }
+{ _: [99.3] }
 ```
 
 _if disabled:_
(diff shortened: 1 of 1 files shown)


bcoe (approved):
LGTM, thank you for the contribution.

bcoe (approved):

Issue #346, opened 2021-01-04
Convert ANSI-C quoted strings to their value


verhovsky:
This makes sense:

```
> yargs.parse('echo --data "\\n"');
{ _: [ 'echo' ], data: '\\n', '$0': '' }
> yargs.parse("echo --data '\\n'");
{ _: [ 'echo' ], data: '\\n', '$0': '' }
```

This could be better:

```
> yargs.parse("echo --data $'\\n'");
{ _: [ 'echo' ], data: "$'\\n'", '$0': '' }
```

In bash, strings of the form `$'string'` are a special type of string:

```bash
$ echo $'some\\ntext'
some\ntext
$ echo 'some\\ntext'
some\\ntext
```

See https://www.gnu.org/software/bash/manual/bash.html#ANSI_002dC-Quoting

It would be helpful if yargs returned the raw value for these types of strings as well, instead of making the user parse it themselves, as this caused (is causing) an issue: https://github.com/NickCarneiro/curlconverter/issues/207

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 50.0 USDC (50.0 USD @ $1.0/USDC)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/yargs/yargs-parser/346/100024881).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/yargs/yargs-parser/346/100024881).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $1,000,591.75 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 265 years, 8 months from now.
Please review their action plans below:


[...]

PR #365, opened 2021-03-11
fix: enable ansi-c escape quoting

jonluca:
Fix for https://github.com/yargs/yargs-parser/issues/346

I think there should be at least a little discussion around this change. 

In general, this seems very safe. it's effectively only fixing bugs. It doesn't go _all_ the way, though - in bash, the escape sequence with \b looks like this:

```
$ echo $'fixx\b me'
fix me
```

where as in this version it'll print out `fixx me`. I'm not sure implementing character deletion makes sense, though - perhaps we could hide actual mutations to the input like this through an option? 

In general this code path is going to be hit so rarely that I don't think it matters _that_ much. Just think people should be aware of it. 

There's also escape sequences like bell which we silently strip out.

### DIFF
--- a/lib/tokenize-arg-string.ts
+++ b/lib/tokenize-arg-string.ts
@@ -1,7 +1,47 @@
+
+/*
+ * This is *only* for bash ANSI-C quoted strings - that is, strings that match the pattern $'some text'
+ *
+ * It's a fairly rare case but leads to the occasional bug. Only implementing the escaped string replacement will most
+ * likely fix the vast majority of these bugs.
+ *
+ * Javascript strings already process some of these correctly (newlines, unicode strings, carriage returns, etc)
+ *
+ * Additional documentation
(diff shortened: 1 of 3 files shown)


verhovsky:
It should keep the escape codes

```sh
$ echo $'fixx\b\b\b\b\b\b me' | xxd
00000000: 6669 7878 0808 0808 0808 206d 650a       fixx...... me.
$ node -e "console.log(process.argv);" $'fixx\b\b\b\b\b\b me'
[ '/usr/bin/node', 'fixx\b\b\b\b\b\b me' ]
```

jonluca:
@verhovsky You're right, I agree. Updated the diff.

jonluca:
@bcoe what's the process for getting the PR landed? Is there anything else you need from me?

jonluca:
Actually @verhovsky https://github.com/yargs/yargs-parser/pull/366/commits seems more thorough and with better tests, let's land that one instead

Issue #302, opened 2020-08-15
Create parser builder pipeline.


Mike111177:
Currently the only way to configure yargs-parser is by the following. (Excuse my mixed syntax)
```js
require('yargs-parser")(args, opts) => Arguments
```
I propose creating a builder pipeline that allows you to preconfigure your opts before running the command by adding an opts function to the Parser interface
```js
export interface Parser {
   //other stuff
   opts: (opts: Partial<Options>) => Parser;
}
```
This has a bunch of advantages:
1. Allows programmer to not keep track of opts variable when used multiple times.
```js
const parse = require('yargs-parser').opts({...})
function myfunc1(){parse("no opts variable here")}
function myfunc2(){parse("no opts variable here")}
```
2. Allows parser to check options ahead of time. This means that if after you create your parser with opts, yargs-parser doesn't have to do any of the setup of validation twice. So if you do `myParser("mystring")` there is no need to 
any of the options pre-processing twice. The first time the args are even referenced on the parser function are like 150 lines of code in: https://github.com/yargs/yargs-parser/blob/5f987aae9c7e7781e54c008b8cff01b210a7a19a/lib/yargs-parser.ts#L203 Since this is not a breaking change you can also now throw an exception on checkConfiguration() errors, since this is a new opt-in feature anyway.

3. Really good interop with ideas from #300 if considered:
```js
//To make a preconfigured bashlike parser
const oldway = require('yargs-parser)
oldway("my command", {opts...})
//Can't throw opts error because that's breaking change
//Warning only happens when parser is used, instead of when options are defined

const parse = require('yargs-parser').opts({...}).bash
//Returns (arg: string, opts?: Partial<Options>) =>Arguments
//Can immediately throw if your options are invalid, faster debugging

parse("myweird option with a literal \\\" in it")
//Is able to skip option validation since it was already done.

parse("myweird option with a literal \\\" in it", {additional_opts})
//Can't skip validation, but original options will still have thrown error if malformed

parse(["my", "bunch", "of", "tokens"])
//Compiler error: bash tokenizer expects string, tokenizing tokens makes no sense
```

These are just a couple of thoughts. Other potential things that could come from this is some of the functionality of yargs being possible in yargs parser, maybe in the future the idea of commands.

bcoe:
> These are just a couple of thoughts. Other potential things that could come from this is some of the functionality of yargs being possible in yargs parser, maybe in the future the idea of commands.

[...]

PR #366, opened 2021-03-15
suport bash's ANSI-C quoted strings

verhovsky:


verhovsky:
Except for numbers and upper/lower case ASCII letters, I actually haven't implemented control code escapes the way bash implements them

```
$ echo -n $'ф' | xxd
00000000: d184                                     ..
$ echo -n $'\cф' | xxd
00000000: 1184                                     ..
$ echo -n $'🟥' | xxd
00000000: f09f 9fa5                                ....
$ echo -n $'\c🟥' | xxd
00000000: 109f 9fa5                                ....
$ echo -n $'^M' | xxd
00000000: 0d                                       .
$ echo -n $'\c^M' | xxd
00000000: 0d                                       .
$ echo -n $'1' | xxd
00000000: 31                                       1
$ echo -n $'\c1' | xxd
00000000: 11                                       .
```

If someone finds it in the bash source code and tells me what it's doing, I can implement it correctly.

verhovsky:
There will be mismatch in parsing because in bash you can go past the limit of the last code point `0x10FFFF`, but you can't do that in node (or Python):

```sh
$ echo -n $'\U10FFFF'  | xxd
00000000: f48f bfbf                                ....
$ echo -n $'\U110000'  | xxd
00000000: f490 8080                                ....
$ node
Welcome to Node.js v14.14.0.
Type ".help" for more information.
> String.fromCodePoint(0x10FFFF)
'􏿿'
> String.fromCodePoint(0x110000)
Uncaught RangeError: Invalid code point 1114112
    at Function.fromCodePoint (<anonymous>)
> 
$ python3
Python 3.9.0+ (default, Oct 19 2020, 09:51:18) 
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "\U0010FFFF"
'\U0010ffff'
>>> "\U00110000"
  File "<stdin>", line 1
    "\U00110000"
                ^
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 0-9: illegal Unicode character
```

jonluca:
@bcoe could we get this merged into upstream?

verhovsky:
@jonluca I still haven't finished control codes, at a minimum it needs to raise an error for the control codes that aren't handled. 

If you'd like, feel free to copy paste parts of this code to curlconverter. Take a look at the source of where Chrome generates the curl command, you don't need to handle everything (like control codes) https://github.com/ChromeDevTools/devtools-frontend/blob/90e5470c94875d4417e5d31e2ab84f9b0682f33c/front_end/network/NetworkLogView.js#L2297-L2326

bcoe:
@verhovsky @jonluca I am supportive of this feature, but since it's breaking and fairly specialized, I'd like for it to be behind the configuration option `--bash-ansi-strings`, then folks who want the behavior for their library just need to turn the feature on in their `package.json`, by adding it to a `yargs` stanza.

test/yargs-parser.cjs
```diff
@@ -3587,6 +3589,121 @@ describe('yargs-parser', function () {
...
+
+    // This passes
+    xit('parses all octal codes in ANSI-C quoted strings like bash', async () => {
+      for (let i = 0; i <= 0o777; i++) {
+        if (i % 256 === 0) { // null character
+          continue
+        }
+        await checkAll('', i.toString(8), 3, 'ascii')
+      }
+    }).timeout(5000)
```
bcoe, line 3663: I'm surprised by the need to extend the timeout, the parsing of strings should be quite fast.

lib/yargs-parser.ts
```diff
@@ -629,6 +632,59 @@ export class YargsParser {
       return value
     }
 
+    // ANSI-C quoted strings are a bash-only feature and have the form $'some text'
+    // https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html
+    function parseAnsiCQuote (str: string): string {
```
bcoe, line 637: Let's pul `lib/string-utils.ts`.

test/yargs-parser.cjs
```diff
@@ -3587,6 +3589,121 @@ describe('yargs-parser', function () {
...
+      for (let i = 1; i <= 0x10FFFF; i++) {
+        if (i >= 0xD800 && i <= 0xDFFF) {
+          continue
+        }
+        await checkAll('U', i.toString(16), 8)
+      }
+
+      // TODO: echo -n $'\U110000' produces f4 90 80 80 but String.fromCodePoint(0x110000)
+      // raises an error in node: Uncaught RangeError: Invalid code point 1114112
+    }).timeout(6000 * 1000)
```
bcoe, line 3691: Likewise, let's move these tests into `test/string-utils.ts`.

bcoe (changes requested):
I'd expect the throughput of the string parsing to be 1000s or 100000s of operations a second, so I don't see why we should need the timeouts added to test cases.

Added a couple additional comments, along with the note I added about putting this behind a feature flag.

test/yargs-parser.cjs
verhovsky: This test runs a bash command through `exec` 511 times. On my desktop it completes in slightly under the default 2 second timeout, but I added some headroom.

verhovsky:
I named it `--parse-bash-ansi-c-strings` instead of `--bash-ansi-strings`. It more verbose but it's more correct, there's no such thing as "ANSI strings". We could name it "--bash-c-strings".

README.md
```diff
@@ -262,6 +262,27 @@ $ node example.js 99.3
...
+* default: `false`
+* key: `parse-bash-ansi-c-strings`
+
+Should arguments that look like [ANSI-C quoted strings](https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html) (a bash-only feature) be treated as such?
+
+_if disabled:_
+
+```console
+> example.js $'hello\\nworld'
+{ _: [ "$'hello\\nworld'" ] }
```
verhovsky, line 276: I wrote this example by hand. how do I generate it properly?

test/string-utils.cjs
```diff
@@ -33,4 +40,102 @@ describe('string-utils', function () {
...
+        value = '0' + value
+        await check("$'\\" + prefix + value + "'", encoding)
+        if (value.toLowerCase() !== value) {
+          await check("$'\\" + prefix + value.toLowerCase() + "'", encoding)
+        }
+      }
+    }
+
+    // Skip exhaustive testing because it takes a long time
+    xit('parses all octal codes in ANSI-C quoted strings like bash', async () => {
```
bcoe, line 81: How do you run these tests? Perhaps you could add an additional `test-exhaustive` script to `package.json`, and we could run them just when we merge to the main branch.

README.md
bcoe: Easiest thing to do might be to clone `yargs` and `yargs-parser`, then do this:

```
cd yargs-parser
npm link .
cd yargs
npm link yargs-parser
// write your example in a file that includes the local copy of yargs.
```

bcoe:
@verhovsky I'm getting back to open source after a break earlier this year. Could I bother you for a rebase 👏 

Working on driving yargs-parser down to zero open PRs.

test/yargs-parser.cjs
verhovsky: I removed these tests, they were more to help me understand how bash does the parsing while I was working on this.

bcoe:
@verhovsky this feature looks quite solid to me, except for a couple a broken test in `optimize`, is this ready to move out of draft?

Thank you for the contribution 👍

bcoe:
@verhovsky friendly nudge, need any help getting this over finish line?

verhovsky:
@bcoe I realized I put this feature in the wrong place. It needs to go in the tokenizer and I have to improve the tokenizer to parse $-prefixed strings first.

yargs has 2 modes of operation, either someone (or a script) calls a JS script through bash, which means bash takes the command strings and parses it (this step parses ANSI-C quoted strings), splitting it into words which then get passed to Node and it exposes that as an array of strings which then get passed to yargs. The other mode is when yargs gets passed a string, presumably this string is a raw bash command that needs to be parsed using bash's rules to build the input array first.

I added my code to work after the first step to operate on the array, which is not the right place for this to go. It needs to happen in [`lib/tokenize-arg-string.ts`](https://github.com/yargs/yargs-parser/blob/54c0d2c9db7647dcae8a3771212ce02974bf5f98/lib/tokenize-arg-string.ts) but I noticed the bash parsing there is quite rudimentary, I thought it would be good to do real parsing. So I marked the PR as draft and started looking into how to parse bash in JavaScript. I had these ideas:

1) bash's parsing is implemented with yacc, here https://git.savannah.gnu.org/cgit/bash.git/tree/parse.y . I thought we could run it with a JS yacc implementation (i.e. [jison](https://github.com/zaach/jison)), but if I understand it correctly, the yacc grammar is used to generate a C program, so you'd still need to rewrite that stuff in JavaScript
2) use someone else's JS bash parser. I only see [js-shell-parse](https://github.com/grncdr/js-shell-parse) and [bash-parser](https://github.com/vorpaljs/bash-parser), both are kinda unmaintained and neither seem to handle ANSI-C quoted strings. What else do they not handle?
3) convert the C code that does shell parsing in a simpler shell such as dash https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/src/parser.c
4) compile parsing logic from either bash or dash to Wasm
5) build on the shell parsing from curl-to-go, which seems a bit better 

You already had basically this conversation in #302

PR #371, opened 2021-04-09
refactor: quote properties used for meta-programming

jkrems:
The library is currently _very_ close to compatible with advanced optimizations done by closure compiler. This is the only issue I found: The properties on the interface may be renamed but the string literals are hard to track back to the property names. Using an explicit enum works around this issue.

I could have used an "normal" enum but this way there should be no observable difference in the non-optimized JavaScript.


bcoe:
@jkrems this looks good to me. Is there a smoke test we can somehow add so that we don't introduce a regression that breaks your use-case?

jkrems:
Testing it can get a bit involved unfortunately. Something like this PR would achieve it though: https://github.com/yargs/yargs-parser/pull/373

bcoe:
@jkrems what order should I land these in, the tests you added seem great to me.

jkrems:
Technically the smoke test PR also has this commit. I wasn't sure where you fall on PR size. I can rebase the smoke test pretty quickly if you want to land the quoting fixes independently.

bcoe (approved):

bcoe:
@jkrems sorry for the slow turnaround on this, have had almost no time for open source this past few months.

I'm working on cleaning up some of these old PRs.

PR #373, opened 2021-04-09
test: add test for optimized output

jkrems:
This adds a test that optimizes the TypeScript using `@tscc/tscc` which combines tsickle and closure compiler with advanced optimizations. The test acts as a rough validation that the library doesn't break when properties are renamed.

Closure compiler will aggressively strip fields or rename them if it thinks they are unused or that it knows all possible use sites. The TypeScript types can already take care of most of the issues but a big exception is meta programming: Dynamic property lookups and -writes need to happen via quoted or bracketed access.


test/tscc/optimized.ts
```diff
@@ -0,0 +1,47 @@
...
+    parser.parse([`--${FLAG_X}=str`]).argv,
+    { _: [], [FLAG_X]: 'str' },
+    'guessed string value')
+  deepEqual(
+    parser.parse([`--no-${FLAG_X}`]).argv,
+    { _: [], [FLAG_X]: false },
+    'guessed boolean negation')
+
+  // Default values for types:
+  deepEqual(
```
jkrems, line 40: This assertion fails before https://github.com/yargs/yargs-parser/pull/371.

test/tscc/lib
```diff
@@ -0,0 +1 @@
+../../lib
```
jkrems, line 1: I needed to add a symlink so that we both get a fresh project root _and_ not have generated closure namespaces with `..` in them.

test/tscc/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "optimized-test",
+  "version": "0.0.0",
+  "dependencies": {
+    "@tscc/tscc": "^0.6.4"
```
jkrems, line 5: The bad news here: It's using a slightly different version of TypeScript (3.9?) right now.

bcoe:
@jkrems awesome, thank you for adding this test 👍

jkrems:
Would you want to combine the two PRs into one? Or is there anything missing on my end to enable landing this?

bcoe (approved):

PR #374, opened 2021-04-16
fix(locale): Turkish camelize and decamelize issues with toLocaleLowerCase/toLocaleUpperCase

sinankeskin:
Hi.

There is an [open issue](https://github.com/stylelint/stylelint/issues/5042) on stylelint for awhile. At first, I suspected immediately  this could be a Turkish locale issue because of the so many Turkish guy complaining about it.

image

If I debug and run `node .\stylelint.js`, I'm seeing some extra flags on console.

image

I investigated deeply and found out that `stylelint` uses a package called `meow` and `meow` uses this package.

I wanna keep short. Because of the Turkish alphabet differences (for ex: lower `ı` = upper `I`, lower `i` = upper `İ`) this issue occurs.

This fix changes all `toLocale...case` functions to `to...case` in order to keep the result same.

Thank you.


metecan (approved):

furcan:
Any news?
Thanks.

bcoe (approved):

bcoe:
@sinankeskin thank you for the contribution 👏 sorry this bug existed for so long.

PR #394, opened 2021-06-20
fix(perf): address slow parse when using unknown-options-as-args

### LINKED ISSUE #386: ReDoS vulnerable when unknown-options-as-args is set
`yargs-parser@20.2.7` appears vulnerable to [regular expression denial-of-service (ReDoS)](https://en.wikipedia.org/wiki/ReDoS) when `unknown-options-as-args` is set to `true` and an attacker can invoke the parser with arguments of the form `----....----a`. See proof-of-concept below:

```javascript
const parse = require('yargs-parser');
for (let i=0; i < 100000; i += 1000)
{
    const s = (new Array(i).fill('-').join('')) + 'a';
    const start = +new Date();
    parse([s], {configuration: {'unknown-options-as-args': true}});
    const end = +new Date();
    console.log(i + '\t' + (end - start));
}
```

See the output of above graphed below, with a power regression overlaid:

image

This appears to be caused by the regexps at `yargs-parser.ts` lines 977-985. The maximum number of characters in an argument passed by process invocation is quite large, so I recommend fixing this by swapping the regular expressions with these safe replacements.

| regexp | replacement|
|-|-|
| `^-+([^=]+?)=[\s\S]*$` | `^-+([^=-]+?)=[\s\S]*$` ||
| `^-+([^=]+?)$` | `^-+([^=-]+?)$` |
| `^-+([^=]+?)-$` | `^-+([^=-]+?)-$` |
| `^-+([^=]+?\d+)$` | `^-+([^=\d-]+?\d+)$` |
| `^-+([^=]+?)\W+.*$` | `^-+([^=\W-]+?)\W+[^\W+]*$` |

bcoe:
@robmcl4 the recommended regex replacements in #386 don't quite work, because of the case of:

`----------foo-bar-snuh`.

Instead I now capture and truncate a large number of leading `-`.

Fixes #386


bcoe:
@robmcl4 feel free to recommend additional test cases, this seemed to catch the examples you provided.

PR #407, opened 2021-07-20
fix: parser should preserve inner quotes

jly36963:
Addresses: https://github.com/yargs/yargs/issues/1324

## Problem

Inner quotes are handled differently by `parse`, depending on the `argsInput` argument's type.

`tokenizeArgString` wraps `argString` with extra quotes if it is a string.
`tokenizeArgString`  does not do the same for `argString`, if it is an array.
The first code block in `processValue` removes quotes from the value. 

### TLDR

If parse was given a string, extra quotes are added (tokenizeArgString) and then removed (processValue).
If parse was given an array, quotes are removed (processValue). This causes loss of inner quotes.

### Example (using tests)

logs for string with inner quotes

```
{ argsInput: 'cmd --foo ""Hello"" --bar ""World"" --baz "":)""' }

during tokenizeArgString 
{ 
  argString: 'cmd --foo ""Hello"" --bar ""World"" --baz "":)""' }
  args: [  'cmd',  '--foo',  '""Hello""',  '--bar',  '""World""',  '--baz', '"":)""'  ]
}

after processValue 
{
  argv: [Object: null prototype] {
    _: [ 'cmd' ],
    foo: '"Hello"',
    bar: '"World"',
    baz: '":)"'
  }
}

result
{ args: { _: [ 'cmd' ], foo: '"Hello"', bar: '"World"', baz: '":)"' } }
```

logs for array with inner quotes

```
{ argsInput: [  'cmd',   '--foo',  '"Hello"', '--bar',  '"World"',   '--baz',   '":)"'  ] }

during tokenizeArgString 
{
  argString: [  'cmd',   '--foo',  '"Hello"', '--bar',  '"World"', '--baz',  '":)"'  ]
  args: [  'cmd',  '--foo',  '"Hello"', '--bar', '"World"', '--baz',  '":)"'  ]
}

after processValue
{
  argv: [Object: null prototype] {
    _: [ 'cmd' ],
    foo: 'Hello',
    bar: 'World',
    baz: ':)' // *** inner quotes missing ***
  }
}
result 
{ args: { _: [ 'cmd' ], foo: 'Hello', bar: 'World', baz: ':)' } }
```

## Solution

I added conditional logic around removing quotes. I feel like my current solution is messy. I would love to hear your thoughts/advice on this. I know you're busy, so no worries if you don't have time. I will try to find a better solution.

There might be follow up work in yargs for handling positionals. The parsed object that gets returned by yargs-parser still has the positional arguments in `argv._`, 

`const parsed = this.#shim.Parser.detailed(...)`

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -557,7 +557,6 @@ export class YargsParser {
         addNewAlias(key, alias)
       }
 
-      console.log({val})
       const value = processValue(key, val, shouldStripQuotes)
       const splitKey = key.split('.')
       setKey(argv, splitKey, value)
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -59,6 +59,9 @@ export class YargsParser {
     // allow a string argument to be passed in rather
     // than an argv array.
     const args = tokenizeArgString(argsInput)
+    // tokenizeArgString adds extra quotes to args if argsInput is a string
+    // only strip those extra quotes in processValue if argsInput is a string
+    const shouldStripQuotes = typeof argsInput === 'string'
 
     // aliases might have transitive relationships, normalize this.
     const aliases = combineAliases(Object.assign(Object.create(null), opts.alias))
@@ -243,7 +246,13 @@ export class YargsParser {
             // nargs format = '--f=monkey washing cat'
             i = eatNargs(i, m[1], args, m[2])
           } else {
-            setArg(m[1], m[2])
+            // only strip quotes if they wouldn't already be removed
+            if (shouldStripQuotes) {
+              setArg(m[1], m[2])
+            } else {
+              setArg(m[1], stripQuotes(m[2]))
+            }
+            
           }
         }
       } else if (arg.match(negatedBoolean) && configuration['boolean-negation']) {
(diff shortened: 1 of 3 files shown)


lib/yargs-parser.ts
```diff
@@ -243,7 +246,13 @@ export class YargsParser {
             // nargs format = '--f=monkey washing cat'
             i = eatNargs(i, m[1], args, m[2])
           } else {
-            setArg(m[1], m[2])
+            // only strip quotes if they wouldn't already be removed
+            if (shouldStripQuotes) {
+              setArg(m[1], m[2])
+            } else {
+              setArg(m[1], stripQuotes(m[2]))
```
bcoe, line 253: Since `setArgs` calls `processValue`, could this be simplified to:

```js
setArg(m[1],, m[2], shouldStripQuotes);
```

bcoe (changes requested):

lib/yargs-parser.ts
jly36963: shouldStripQuotes should always be forwarded by `setArg` to `processValue`, except at this line. In this invocation, shouldStripQuotes should be true. 

I'm not a fan of the solution I have [here](https://github.com/yargs/yargs-parser/blob/5ed073d480242be7c4959487ca12f39b4401e30a/lib/yargs-parser.ts#L546)

I try to keep code purely functional (where possible). In order to avoid reading global variables in setArg, I dynamically set the default value for `shouldStripQuotes`. I don't particularly like that pattern, so I'm down to change it.

bcoe (approved):

bcoe:
@jly36963 thanks for the patch 👌

PR #416, opened 2021-10-10
fix: esm json import

jly36963:
Addresses: https://github.com/yargs/yargs/issues/2040

## Problem

The esm require for json files wasn't working properly. I offered a workaround in a comment on the issue above (passing a custom ConfigCallback to `yargs.config()`), but the underlying issue was the way yargs-parser handles the importing of the json file.

## Solution

`readFileSync(path, 'utf8')` reads the json file as a string, but it needs to be a parsed object. To make it behave the same way as the cjs `require`, there are two easy solutions.

```js
// use module createRequire
import { createRequire } from "module";
const esmRequire = createRequire(import.meta.url);

const config = esmRequire(path);
```

```js
// parse the json string
const config = JSON.parse(readFileSync(path, 'utf8'));
```

I chose the `createRequire` solution, which required me to change the modules from es2015 to es2020 in tsconfig. I believe it would also require dropping support for Node 10, as `createRequire` was implemented in node 12. If that's a deal-breaker, I'm cool with switching to the `JSON.parse` solution instead.

## Note

It looks like optimize and typescript are at odds now. Because of the TS 4.4 breaking change, I need to add types to the catch block errors, but optimize doesn't like it.


jly36963:
@bcoe sounds good

jly36963:
Let me know what else I can do on https://github.com/yargs/yargs-parser/pull/421

bcoe:
@jkrems thank you for the review.

bcoe (approved):
I like where this work landed 👍

PR #432, opened 2022-01-25
fix: return deno env object

johanholmerin:
Assume the environment object should actually be returned

Fixes https://github.com/yargs/yargs/issues/2136


bcoe (approved):

bcoe:
@johanholmerin at a glance patch looks good, but seems to not be working as expected in Deno tests.

bcoe (approved):

bcoe:
@johanholmerin thank you for the contribution, this should address https://github.com/yargs/yargs/issues/2136 I believe.

webbertakken:
Created a PR ^ that will update `yargs-parser` in `yargs`, so that `.env()` can effectively be used again in Deno.

PR #439, opened 2022-04-20
Run linting on `*.mjs` files as well

voxpelli:
Will highlight the linting issues we encountered in `standard`: https://github.com/standard/standard/issues/1777#issuecomment-1103931191

Also: Since you're not running stock `standard` I guess its best if we remove you from our list of modules to regression test?

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
   "type": "module",
   "module": "./build/lib/index.js",
   "scripts": {
-    "check": "standardx '**/*.ts' && standardx '**/*.js' && standardx '**/*.cjs'",
-    "fix": "standardx --fix '**/*.ts' && standardx --fix '**/*.js' && standardx --fix '**/*.cjs'",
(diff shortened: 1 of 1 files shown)


shadowspawn:
Thanks @voxpelli 

Given this project PRs are a bit backlogged, I don't this is a good regression test currently. I don't know whether using `standardx` affects suitability for regression test.

shadowspawn:
The line that breaks `standardx` with `yargs-parser.mjs ` and (old) current dependencies is:

```
  const __dirname = path.dirname(fileURLToPath(import.meta.url))
```

```
standardx: Use JavaScript Standard Style (tweaked by standardx) (https://github.com/standard/standardx)
  /Users/john/Documents/Sandpits/yargs/yargs-parser/my-fork/test/yargs-parser.mjs:10:48: Parsing error: Unexpected token import
```

Commenting that out, there are about five sorts of warnings.

PR #443, opened 2022-06-16
Allow the browser build to be imported

TechplexEngineer:
Adding this small change will allow cloudflare worker users to easily use the yargs-parser as a library.


bcoe:
@TechplexEngineer thank you for the contribution, that would be a great side-effect of ESM support, if we could run in a worker environment 😄 

Let me know if there are any more changes needed.

PR #450, opened 2022-07-13
fix: node version check now uses process.versions.node

### LINKED ISSUE #447: Version number check fails on Bun
I decided to give Bun a try by running the Mocha test suite of one of my packages.  The very first failure was in `yargs-parser` like so:

```
$ bun run ./node_modules/mocha/bin/mocha.js test-partials.js test-parse.js test-process.js 
986 |     ? Number(process.env.YARGS_MIN_NODE_VERSION)
987 |     : 10;
988 | if (process && process.version) {
989 |     const major = Number(process.version.match(/v([^.]+)/)[1]);
990 |     if (major < minNodeVersion) {
991 |         throw Error(`yargs parser supports a minimum Node.js version of ${minNodeVersion}. Read our version support policy: https://github.com/yargs/yargs-parser#supported-nodejs-versions`);
                  ^
 error: yargs parser supports a minimum Node.js version of 10. Read our version support policy: https://github.com/yargs/yargs-parser#supported-nodejs-versions
      at /home/david/Projects/akasharender/mahabhuta/test/node_modules/yargs-parser/build/index.cjs:991:14
      at bun:wrap:1:16354
      at /home/david/Projects/akasharender/mahabhuta/test/node_modules/mocha/lib/cli/options.js:12:28
      at bun:wrap:1:16354
      at /home/david/Projects/akasharender/mahabhuta/test/node_modules/mocha/bin/mocha.js:13:30
```

Clearly, Bun isn't going to give a value for `process.version` that is compatible with Node.js versions.  With this script

```js
console.log(process.version);

console.log(process);
```

I get this output:

```
$ bun run pv.js 
v0.1.2
{ ... }
```

For reference: https://github.com/Jarred-Sumner/bun/issues/462

paperdave:
Closes #447 by using `process.versions.node` for testing the version. The only difference between these values is the `v` prefix, so it's not as trivial as swapping a variable. Also, just in case `process.versions` is somehow not set, this falls back to previous behavior.

With this change in place, Bun should be able to import yargs.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -19,8 +19,9 @@ import { readFileSync } from 'fs'
 const minNodeVersion = (process && process.env && process.env.YARGS_MIN_NODE_VERSION)
   ? Number(process.env.YARGS_MIN_NODE_VERSION)
   : 12
-if (process && process.version) {
-  const major = Number(process.version.match(/v([^.]+)/)![1])
+const nodeVersion = process && (process.versions ? process.versions.node : process.version.slice(1))
+if (nodeVersion) {
+  const major = Number(nodeVersion.match(/^([^.]+)/)![1])
   if (major < minNodeVersion) {
     throw Error(`yargs parser supports a minimum Node.js version of ${minNodeVersion}. Read our version support policy: https://github.com/yargs/yargs-parser#supported-nodejs-versions`)
   }
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -19,7 +19,7 @@ import { readFileSync } from 'fs'
 const minNodeVersion = (process && process.env && process.env.YARGS_MIN_NODE_VERSION)
   ? Number(process.env.YARGS_MIN_NODE_VERSION)
   : 12
-const nodeVersion = process && (process.versions ? process.versions.node : process.version.slice(1))
+const nodeVersion = process?.versions?.node || process?.version?.slice(1)
 if (nodeVersion) {
   const major = Number(nodeVersion.match(/^([^.]+)/)![1])
   if (major < minNodeVersion) {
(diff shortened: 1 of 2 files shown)


lib/index.ts
```diff
@@ -19,8 +19,9 @@ import { readFileSync } from 'fs'
 const minNodeVersion = (process && process.env && process.env.YARGS_MIN_NODE_VERSION)
   ? Number(process.env.YARGS_MIN_NODE_VERSION)
   : 12
-if (process && process.version) {
-  const major = Number(process.version.match(/v([^.]+)/)![1])
+const nodeVersion = process && (process.versions ? process.versions.node : process.version.slice(1))
```
bcoe, line 22: Could you do:

```
const nodeVersion = process?.versions?.node && process.version.slice(1);
```

To simplify the logic slightly?

bcoe (changes requested):
Fix seem reasonable to me, left a small nit.

lib/index.ts
davecaruso: this should be an `||` but yes the logic could be simplified. i wasn't sure if i was able to use `?.` in the code.

lib/index.ts
```diff
@@ -19,8 +19,9 @@ import { readFileSync } from 'fs'
 const minNodeVersion = (process && process.env && process.env.YARGS_MIN_NODE_VERSION)
   ? Number(process.env.YARGS_MIN_NODE_VERSION)
   : 12
-if (process && process.version) {
-  const major = Number(process.version.match(/v([^.]+)/)![1])
+const nodeVersion = process?.versions?.node || process?.version?.slice(1)
```
bcoe, line 22:
```suggestion
const nodeVersion = process?.versions?.node ?? process?.version?.slice(1)
```

bcoe:
@davecaruso I believe there's been a small dip in coverage because of the newline added, please feel free to update the file here:

https://github.com/yargs/yargs-parser/blob/main/.nycrc

On your branch, to bring the threshold down to 97.

bcoe (approved):

bcoe:
@paperdave thank you for the contribution 👌

PR #457, opened 2022-08-06
Implemented option to delay treating non-flags as haltable

Toberumono:
This is a temporary solution that allows for most of the halt-at-non-option functionality to work with subcommands and is based on some problems revealed by #438 and #453. Ideally this will be replaced in the future by having yargs-parser support command words (see the later discussion in #456 for details).

The flag name is open for changes. I'm terrible at naming things.


bcoe:
@Toberumono these test cases are useful, but I'm hesitant to introduce the config option `wait-for-first-flag-before-filtering-options`.

This feels very much like a case of https://xkcd.com/1172/, and I'd rather figure out an approach to solving this problem that's more intuitive.

bcoe:
@Toberumono I left some comments on #453, I think I need to better understand your use case before I have an opinion about a fix (or adding an additional config setting).

PR #505, opened 2025-06-18
feat: Implement robust browser path resolution

mitsuki31:
This PR addresses the long-standing `TODO` item in `browser.js` by integrating `path-browserify` to provide robust path normalization and resolution capabilities for browser environments.

Previously, `browser.js` relied on no-op implementations for `normalize` and `resolve`.
https://github.com/yargs/yargs-parser/blob/553c808cb0a8730980cacd23b8b8e3038ab7c61f/browser.js#L7-L16

After integrating `path-browserify`, the `normalize` and `resolve` functions have been updated to directly reference `path-browserify.normalize` and `path-browserify.resolve`, respectively.
https://github.com/yargs/yargs-parser/blob/08254e37686f20169a5ee0285b9a6c1deb3c2c0d/browser.js#L6-L18

### Integration of `path-browserify` as a Vendored Module
- The `path-browserify` library, which offers a browser-compatible API mirroring Node.js's `path` module (specifically Node.js v10.3), has been included directly within the `vendor/path-browserify` directory.
- This package was fetched directly from the official npm registry (`path-browserify-1.0.1.tgz`) using `wget`.
  ```bash
  wget https://registry.npmjs.org/path-browserify/-/path-browserify-1.0.1.tgz
  ```
  > I also removed the `path-browserify`'s "test" directory to ensure the package size as small as possible.
- To ensure compatibility with browser ES module imports, the `path-browserify`'s primary entry point (`index.js`) has been slightly modified to use ES module exports (`export default posix;`).
  https://github.com/yargs/yargs-parser/blob/08254e37686f20169a5ee0285b9a6c1deb3c2c0d/vendor/path-browserify/index.js#L527-L529

### `browser.js` Path Function Update
- The `normalize` and `resolve` functions within the `YargsParser` constructor in `browser.js` now directly reference `path-browserify.normalize` and `path-browserify.resolve`, leveraging the new capabilities.

### Build and Linting Adjustments
- `.eslintignore` file has been reconfigured to ignore the `vendor` directory during linting.
- The project's bundling configuration has been updated to explicitly include the contents of the `vendor` directory, ensuring `path-browserify` is correctly bundled into the final distribution.

## Justification for Dependency Management Strategy (Manual Vendoring)

The decision to manually vendor `path-browserify` (by fetching the `1.0.1.tgz` directly from the npm registry) rather than adding it as a standard `npm` dependency was made for the following reasons:

* The `path-browserify` package, while functionally complete for this purpose, has not received updates on npm for a number of years. Manually vendoring allows for precise control over the exact version included in `yargs-parser`'s browser bundle, isolating it from potential future `npm` dependency changes or deprecations.
* This approach allows for the necessary modification of `path-browserify`'s `index.js` to expose ES module exports, which is crucial for `browser.js`'s current import structure.

## Testing

```html
<!DOCTYPE html>
<body>
  <script type="module">
    import yargs from "./browser.js";

    const argv = yargs(
      '--testPath=/foo/bar/../bazz//a/filename.txt --testPath-2=a/b/../f/../../foo/file.txt',
      {
        normalize: ['testPath', 'testPath-2']
      }
    )
    console.log(argv.testPath)
    console.log(argv.testPath2)
    console.log(argv.testPath === '/foo/bazz/a/filename.txt')  // Should return true
    console.log(argv.testPath2 === 'foo/file.txt')             // Should return true
  </script>
</body>
```

---

I am keen to discuss these considerations and explore any preferred alternative approaches to integrating Browser version of Node.js `path` module.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 build/
+vendor/
 test/tscc/
--- a/browser.js
+++ b/browser.js
@@ -1,14 +1,16 @@
 // Main entrypoint for ESM web browser environments. Avoids using Node.js
 // specific libraries, such as "path".
 //
-// TODO: figure out reasonable web equivalents for "resolve", "normalize", etc.
 import { camelCase, decamelize, looksLikeNumber } from './build/lib/string-utils.js'
 import { YargsParser } from './build/lib/yargs-parser.js'
+// Using `path-browserify` package from local dependency
+import path from './vendor/path-browserify/index.js';
+
 const parser = new YargsParser({
   cwd: () => { return '' },
   format: (str, arg) => { return str.replace('%s', arg) },
-  normalize: (str) => { return str },
-  resolve: (str) => { return str },
+  normalize: path.normalize,  // Use `path-browserify`
+  resolve: path.resolve,    // Use `path-browserify`
   require: () => {
     throw Error('loading config from files not currently supported in browser')
   },
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
   "files": [
     "browser.js",
     "build",
+    "vendor",
     "!*.d.ts",
     "!*.d.cts"
   ],
--- a/vendor/path-browserify/.github/FUNDING.yml
+++ b/vendor/path-browserify/.github/FUNDING.yml
@@ -1,0 +1,12 @@
+# These are supported funding model platforms
+
+github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
+ko_fi: # Replace with a single Ko-fi username
+tidelift: npm/path-browserify
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+liberapay: # Replace with a single Liberapay username
+issuehunt: # Replace with a single IssueHunt username
+otechie: # Replace with a single Otechie username
+custom: # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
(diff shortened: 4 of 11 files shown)


shadowspawn:
To understand a bit more about what this would allow. Is `opts.normalize` an option you want to use yourself in the browser? What is the use case that it helps with?

shadowspawn:
Interesting to see you found a module for this very purpose.

My initial thought is that vendoring in `path-browserify` is more overhead than we want for a feature that no one has asked about in a less common runtime. (i.e. I don't see any issues in yargs or yargs-browser mentioning `normalize` and `browser`.) 

I was interested to see that according to the path-browserify README, it was included by default in some bundlers. If that is still a common pattern, a light weight approach might be to look for an available implementation supplied by the bundler. 

> You usually do not have to install path-browserify yourself! If your code runs in Node.js, path is built in. If your code runs in the browser, bundlers like [browserify](https://github.com/browserify/browserify) or [webpack](https://github.com/webpack/webpack) include the path-browserify module by default.

Another approach would be to make it possible to supply own implementations so end-user can implement it if needed.

It would be much smaller to only vendor the relevant routines.

I am only one maintainer! I suggest you wait for feedback from other maintainers or end users before doing more work on this.

mitsuki31:
Thank you for the clear guidance.

I understand your concerns about the overhead of the current vendoring approach for `path-browserify` and the current demand for this functionality in the browser runtime. To keep the queue clear, I'm glad to close this PR if the present vendoring approach doesn't seem to be the preferred direction for this project.

shadowspawn:
There have not been any user requests for this functionality. Not planning to vendor-in.

Thanks for the discussion.

PR #515, opened 2025-09-04
docs: remove greedy arrays note

mcecode:
Feel free to close if I missed something and it was changed, but as far as I can tell, the default wasn't changed in version 18. So, I removed the note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,8 +346,6 @@ $ node example --arr 1 2
 { _: [2], arr: [1] }
 ```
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
Thanks. v18 was very focused on the switch to esm-only.

Per #249 I think the comment is still relevant, but it would now be considered for v19 since it was not considered for v18.

mcecode:
Closing in favor of #516

PR #523, opened 2026-02-07
fix: improve boolean env var coercion and add NO_ prefix negation support

### LINKED ISSUE #2501: Boolean arguments read via environment variables do not support no-prefixed variants
Given the following script:

```main.js
#!/usr/bin/env node
import yargs from 'yargs';
import { hideBin } from 'yargs/helpers';
const argv = yargs(
  hideBin(process.argv)
).env(
 "MY_CMD"
).option(
 'doThing',
 {type: 'boolean', default: true}
).parse()

console.log(`DO THING? ${argv.doThing}`);
```

My expectation would be that `MY_CMD_DO_THING`'s environment variable would map to booleans in "the right way". I'd also expect for `MY_CMD_NO_DO_THING` to map to `--no-do-thing`'s behavior.

Instead I've experienced the following:

```
+ bun run ./main.js
DO THING? true // expected: default is true
+ MY_CMD_DO_THING=1
+ bun run ./main.js
DO THING? false // unexpected: expected true by "1" being "truth"-y
+ MY_CMD_DO_THING=0
+ bun run ./main.js
DO THING? false // expected: expected false by "0" being "false"-y
+ MY_CMD_DO_THING=true
+ bun run ./main.js
DO THING? true // seems fine
+ MY_CMD_DO_THING=false
+ bun run ./main.js
DO THING? false // seems fine
+ MY_CMD_NO_DO_THING=1
+ bun run ./main.js
DO THING? true  // unexpected: expected false by no-do-thing + 1 being "truth"-y
+ MY_CMD_NO_DO_THING=0
+ bun run ./main.js
DO THING? true // guess this is right
+ MY_CMD_NO_DO_THING=true
+ bun run ./main.js
DO THING? true. // unexpected: expected false by no-do-thing + true being truthy
+ MY_CMD_NO_DO_THING=false
+ bun run ./main.js
DO THING? true
```

I _think_ `NO_DO_THING` is just not looked at in its entirety, which isn't the end of the world, but seems to lack some notion of parity with the rest of the negation behavior.

But for CLI applications, having `1` or `0` represent `true` or `false` feels like something that makes sense. Here `1` turning into a false-y result is very surprising to me.

kaigritun:
## Summary

Fixes yargs/yargs#2501

This PR addresses two related issues with how boolean options are handled when set via environment variables:

### Issue 1: Boolean string coercion only accepts 'true'

Previously, when a boolean option was set via environment variable, only the exact string `'true'` was recognized as truthy. This caused unexpected behavior:

```
MY_CMD_DO_THING=1     # became false (expected: true)
MY_CMD_DO_THING=yes   # became false (expected: true)
MY_CMD_DO_THING=true  # worked correctly
```

**Fix:** Boolean string coercion now accepts common truthy/falsy values:
- Truthy: `'true'`, `'1'`, `'yes'`, `'on'`
- Falsy: `'false'`, `'0'`, `'no'`, `'off'`

### Issue 2: NO_ prefix not recognized for boolean negation

The CLI supports `--no-doThing` for boolean negation, but the equivalent env var `MY_CMD_NO_DO_THING` was not recognized as negation - it was treated as a separate option.

**Fix:** Environment variables with `NO_` prefix (after the env prefix) are now treated as boolean negation:
- `MY_CMD_NO_DO_THING=true` → `doThing: false`
- `MY_CMD_NO_DO_THING=false` → `doThing: true` (double negation)

The implementation respects the `'negation-prefix'` and `'boolean-negation'` configuration options.

## Testing

Added 7 new test cases covering:
- Boolean env var with `'1'` value
- Boolean env var with `'0'` value
- Boolean env var with `'yes'` value
- Boolean env var with `'no'` value
- `NO_` prefix with `'true'` value
- `NO_` prefix with `'1'` value
- `NO_` prefix with `'false'` value (double negation)

All existing tests continue to pass (369 total).

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -617,7 +617,18 @@ export class YargsParser {
 
       // handle parsing boolean arguments --foo=true --bar false.
       if (checkAllAliases(key, flags.bools) || checkAllAliases(key, flags.counts)) {
-        if (typeof val === 'string') val = val === 'true'
+        if (typeof val === 'string') {
+          const lower = val.toLowerCase()
+          // Support common truthy/falsy string representations
+          if (lower === 'true' || lower === '1' || lower === 'yes' || lower === 'on') {
+            val = true
+          } else if (lower === 'false' || lower === '0' || lower === 'no' || lower === 'off') {
+            val = false
+          } else {
+            // For other strings, treat non-empty as truthy (backwards compat edge case)
+            val = val === 'true'
+          }
+        }
       }
 
       let value = Array.isArray(val)
@@ -728,19 +739,48 @@ export class YargsParser {
       if (typeof envPrefix === 'undefined') return
 
       const prefix = typeof envPrefix === 'string' ? envPrefix : ''
+      // Convert negation-prefix to env var format (e.g., 'no-' -> 'NO_')
+      const negationPrefix = (configuration['negation-prefix'] || 'no-').toUpperCase().replace(/-/g, '_')
       const env = mixin.env()
       Object.keys(env).forEach(function (envVar) {
         if (prefix === '' || envVar.lastIndexOf(prefix, 0) === 0) {
           // get array of nested keys and convert them to camel case
+          let isNegated = false
           const keys = envVar.split('__').map(function (key, i) {
             if (i === 0) {
               key = key.substring(prefix.length)
+              // Check for negation prefix (e.g., NO_DO_THING -> doThing with negation)
+              if (configuration['boolean-negation'] && key.lastIndexOf(negationPrefix, 0) === 0) {
+                key = key.substring(negationPrefix.length)
+                isNegated = true
+              }
             }
             return camelCase(key)
           })
 
           if (((configOnly && flags.configs[keys.join('.')]) || !configOnly) && !hasKey(argv, keys)) {
-            setArg(keys.join('.'), env[envVar])
+            let value: any = env[envVar]
+            // Handle negated boolean env vars
+            if (isNegated && checkAllAliases(keys.join('.'), flags.bools)) {
+              // For negated booleans: NO_FOO=true means foo=false, NO_FOO=false means foo=true
+              const lower = String(value).toLowerCase()
+              if (lower === 'true' || lower === '1' || lower === 'yes' || lower === 'on') {
+                value = false
+              } else if (lower === 'false' || lower === '0' || lower === 'no' || lower === 'off') {
+                value = true
+              } else {
+                value = false // Default negated to false for non-standard values
+              }
+              setKey(argv, keys, value)
+              // Also set aliases
+              if (flags.aliases[keys.join('.')]) {
+                flags.aliases[keys.join('.')].forEach(function (x) {
+                  setKey(argv, x.split('.'), value)
+                })
+              }
+            } else {
+              setArg(keys.join('.'), value)
+            }
           }
         }
       })
--- a/test/yargs-parser.mjs
+++ b/test/yargs-parser.mjs
@@ -2306,6 +2306,82 @@ describe('yargs-parser', function () {
         bar: 'bar'
       })
     })
+
+    it('should coerce boolean env var "1" to true', function () {
+      process.env.TEST_BOOL_DO_THING = '1'
+      const result = parser([], {
+        envPrefix: 'TEST_BOOL',
+        boolean: ['doThing']
+      })
+      result.doThing.should.equal(true)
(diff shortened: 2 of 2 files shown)


shadowspawn:
Interesting. I hadn't commented on what I thought behaviour should be in https://github.com/yargs/yargs/issues/2501#issuecomment-3864531225, just confirmed how it currently worked. 

I have been doing some research. For `NO_` negation...

I do not think we should support `NO_` negation. I do see the appeal for parity with `--no`, but boolean options are detected by their presence whereas we are already parsing environment variable values. We have one option `--foo` which can be negated with `--no-foo`. We have one environment variable `FOO` which can be negated with `FOO=false`. Negated environment variables are reasonably rare in practice (with legacy `NO_COLOR` an example).

Common use of options:
```console
mycmd --print
mycmd --no-print
```

Common (and clear) use of environment variables which use values:
```console
MY_CMD_PRINT=true
MY_CMD_PRINT=false
```

Negated variables with a value are less clear. (I personally avoid creating variables that start with a "no".)
```console
MY_CMD_NO_PRINT=true
MY_CMD_NO_PRINT=false # yuck!
```

And lastly, adding a negation introduces a potential conflict, as now two variables instead of one. Which one wins? On command-line the last one wins, but that does not work for environment variables.
```console
MY_CMD_PRINT=true
MY_CMD_NO_PRINT=true
```

shadowspawn:
I think the 1/0 recognition for environment variables covers a common usage and expectation. I am not so keen on the yes/no and off/on. I have not come across those often in the wild. Are you aware of utilities that support those, or where they are commonly used?

kaigritun:
Good question! The 1/0 is definitely the most common. For yes/no and on/off:

- **yes/no**: Common in interactive prompts (e.g., `npm init` style tools) and some config systems like Apache httpd (`KeepAlive On/Off`), MySQL (`query_cache_type=ON`)
- **on/off**: Used in git config (`core.autocrlf=true` but also accepts `on`), PHP (`display_errors = On`), and various INI-style configs

That said, if you'd prefer to keep it minimal and just support 1/0 (plus the existing true/false), I'm happy to trim it down. The core fix for the coverage issue is separate from the extended value recognition.

lib/yargs-parser.ts
```diff
@@ -617,7 +617,18 @@ export class YargsParser {
...
+        if (typeof val === 'string') {
+          const lower = val.toLowerCase()
+          // Support common truthy/falsy string representations
+          if (lower === 'true' || lower === '1' || lower === 'yes' || lower === 'on') {
+            val = true
+          } else if (lower === 'false' || lower === '0' || lower === 'no' || lower === 'off') {
+            val = false
+          } else {
+            // For other strings, treat non-empty as truthy (backwards compat edge case)
+            val = val === 'true'
```
shadowspawn, line 629:
```suggestion
            // For other strings, treat as false (backwards compat edge case)
            val = false
```

lib/yargs-parser.ts
shadowspawn, line 612: Making changes in `processValue` affects more than just environment variables. I think that is potentially ok, somewhat of an edge case, and maintains compatibility between ENV and other cases.

I _think_ it just affects `--foo=VALUE` when `foo` is boolean.

shadowspawn:
> That said, if you'd prefer to keep it minimal and just support 1/0 (plus the existing true/false), I'm happy to trim it down.

Thanks for examples. Keep them in for now, and I'll keep thinking. 🤔 😄

MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

shadowspawn:
Thanks @MichaelDeBoey , I had seen the story but not connected it with this PR!

mdekstrand:
Just found this PR chasing down Yargs env var support and why `PROG_VAR=1` wasn't working for me. I would love to see yargs support more flexibility on environment variable inputs for booleans.

If adding “reasonably complete support for common truthy/falsey values” is a reasonable goal here, I'd suggest also accepting single-letter `t`/`y`/`f`/`n`.

shadowspawn:
Thanks for feedback @mdekstrand 

Do you use systems where `t`/`y`/`f`/`n` are commonly used in environment variables? (Or more filling out the possibilities?)

mdekstrand:
Mostly filling out the possibilities, but I know I've seen y/n show up somewhere (I don't remember off-hand what program(s) that was, though).

PR #533, opened 2026-06-19
fix: array type flags (number/boolean/string) should respect their value

### LINKED ISSUE #415: Existence of ‘number’ Key in `Options.arrays` Changes Parsing
The [documentation](https://github.com/yargs/yargs-parser#parserargs-opts) says to set `numbers: true` to parse option-array values as numbers. However, this value can be anything other than `true` and yargs will still use number parsing. This is particularly confusing when trying to set `numbers: false` or `numbers: undefined`. 

```
> const parser = require('yargs-parser')
undefined
> parser('--guk dog cat', {array: {key: 'guk'}})
{ _: [], guk: [ 'dog', 'cat' ] }
> parser('--guk dog cat', {array: {key: 'guk', number: false}})
{ _: [], guk: [ NaN, NaN ] }
> parser('--guk dog cat', {array: {key: 'guk', number: undefined}})
{ _: [], guk: [ NaN, NaN ] } 
```

This applies to `boolean` as well.

chatman-media:
Fixes #415.

## Problem

When an `array` option is configured with an object, the parser decides whether to coerce its values to `number`/`boolean`/`string` based on the **presence** of the corresponding key, ignoring its value. As a result a *falsy* type flag still enables coercion:

```js
const parser = require("yargs-parser")

parser("--foo dog cat", { array: { key: "foo", number: false } })
// actual:   { _: [], foo: [ null, null ] }   (NaN on older versions)
// expected: { _: [], foo: [ "dog", "cat" ] }

parser("--foo dog cat", { array: { key: "foo", number: undefined } })
// same wrong result
```

The README documents enabling coercion with `number: true`, and the type definition declares these properties as `boolean?`, so `false`/`undefined` should behave as if the flag were absent. The same bug affects `boolean: false` (values wrongly coerced to booleans) and `string: false` (numbers wrongly kept as strings).

## Cause

In `lib/yargs-parser.ts`, the assignment was derived from `Object.keys(opt)`, which only checks key existence:

```js
const assignment = Object.keys(opt).map(function (key) {
  const arrayFlagKeys = { boolean: "bools", string: "strings", number: "numbers" }
  return arrayFlagKeys[key]
}).filter(Boolean).pop()
```

## Fix

Only enable a coercion flag when its value is strictly `=== true`. (Also guards the object-vs-string `opt` case explicitly, instead of relying on a string`s numeric character indices never matching a flag name.)

## Tests

Added four tests covering `number: false`, `number: undefined`, `boolean: false`, and `string: false`. All four fail on `main` and pass with the fix; the full suite (`npm test`) stays green (366 passing) with 100% coverage on `yargs-parser.ts`, and `npm run check` (gts lint) is clean.

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -115,14 +115,17 @@ export class YargsParser {
       const key = typeof opt === 'object' ? opt.key : opt
 
       // assign to flags[bools|strings|numbers]
-      const assignment: ArrayFlagsKey | undefined = Object.keys(opt).map(function (key) {
-        const arrayFlagKeys: Record<string, ArrayFlagsKey> = {
-          boolean: 'bools',
-          string: 'strings',
-          number: 'numbers'
-        }
-        return arrayFlagKeys[key]
-      }).filter(Boolean).pop()
+      const arrayFlagKeys: Record<string, ArrayFlagsKey> = {
+        boolean: 'bools',
+        string: 'strings',
+        number: 'numbers'
+      }
+      const assignment: ArrayFlagsKey | undefined = typeof opt === 'object'
+        ? Object.keys(arrayFlagKeys)
+          .filter((flagKey) => (opt as Record<string, unknown>)[flagKey] === true)
+          .map((flagKey) => arrayFlagKeys[flagKey])
+          .pop()
+        : undefined
 
       // assign key to be coerced
       if (assignment) {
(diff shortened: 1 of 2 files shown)


shadowspawn:
(Will be a while before I review this. Both the original code and replacement code are somewhat opaque, and other issues waiting.)

Issue #506, opened 2025-06-26
Multiple arguments are incremented if they are equal to 1.


dennis-grunert:
Yargs version used: `18.0.0`

## How to reproduce

```javascript
const parser = require('yargs-parser')
parser("-x 3 -x 1")
```

returns

```javascript
{ _: [], x: 4 }
```

i.e., it wrongfully calculated `3 + 1 = 4`.

Expected would be
```javascript
{ _: [], x: [ 3, 1 ] }
```

The option `"duplicate-arguments-array": false` still results in `4`.

If the order is changed, i.e.,
```javascript
const parser = require('yargs-parser')
parser("-x 1 -x 3")
```
then the result is correct.

## Bug Cause

The increment happens in `yargs-parser.ts` -> `setKey` at [this code line](https://github.com/yargs/yargs-parser/blob/553c808cb0a8730980cacd23b8b8e3038ab7c61f/lib/yargs-parser.ts#L853):
```javascript
if (value === increment()) {
   o[key] = increment(o[key])
} 
```

`increment() === 1`, i.e., it increments `o[key]` each time we have any argument which equals to `1`.

The [increment function](https://github.com/yargs/yargs-parser/blob/553c808cb0a8730980cacd23b8b8e3038ab7c61f/lib/yargs-parser.ts#L1110) has a comment on top stating that it "should only be called when a count is given as an arg". Obviously, we have a number and not a count here.

## Workaround

A workaround in `yargs/yargs` consists in setting `type="string"`.

shadowspawn:
The bug got introduced in the conversion to TypeScript. The function itself was being used as a magic value previously, and quite different than testing the function return value.

https://github.com/yargs/yargs-parser/blob/d9284ec1466d9934b7be8039db27d73dcb4de8f8/index.js#L519

https://github.com/yargs/yargs-parser/blob/d9284ec1466d9934b7be8039db27d73dcb4de8f8/index.js#L727

PR #534, opened 2026-07-06
test: add regression for non-count option with value 1

ZhouYinLong-lab:
## What changed
Added a regression test for non-count options that receive a value of `1`.

## Why
When passing `-x 3 -x 1`, the parser incorrectly treats the `1` as a count increment, returning `x: 4` instead of `x: [3, 1]`. The root cause is in `setKey()` where `value === increment()` (which is `1`) gates the increment logic, incorrectly matching actual values of `1`.

## Verification
```
parser("-x 3 -x 1")
// Expected: x: [3, 1]
// Actual:   x: 4  ← bug: 3 + 1 = 4
```
The test currently fails as expected.

## Related issue
Refs #506

### DIFF
--- a/test/yargs-parser.mjs
+++ b/test/yargs-parser.mjs
@@ -1529,6 +1529,13 @@ describe('yargs-parser', function () {
     })
   })
 
+
+  it('should not increment non-count options when value equals 1', function () {
+    const result = parser('-x 3 -x 1')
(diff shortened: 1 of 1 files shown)


shadowspawn:
1) This looks like AI slop
2) These is already an open PR #529

PR #541, opened 2026-08-14
fix: parse short option with trailing equals and empty value (-o=) as empty string

### LINKED ISSUE #367: Using an equal sign with short/alias options and an empty string
Short options do not seem to work when using an equal sign and an empty string.

Example:

```js
// yargs.js
const yargs = require('yargs')
console.log(yargs.option('opt', { alias: 'o' }).parse())
```

```
$ node yargs.js -o ""
{ _: [], o: '', opt: '', '$0': 'yargs.js' }
$ node yargs.js -o=""
{ _: [], o: true, opt: true, '=': true, '$0': 'yargs.js' }
```

Note that this works correctly when not using an empty string:

```
$ node yargs.js -o "value"
{ _: [], o: 'value', opt: 'value', '$0': 'yargs.js' }
$ node yargs.js -o="value"
{ _: [], o: 'value', opt: 'value', '$0': 'yargs.js' }
```

Using yargs `16.2.0` on Node `15.6.0`.

youdie006:
Fixes #367

## Problem

A short option written with a trailing `=` and no value mis-parses. The `=` is treated as a standalone key and the option becomes boolean `true`:

    parser(['-o='], { alias: { o: 'opt' } })
    // actual:   { _: [], o: true, opt: true, '=': true }
    // expected: { _: [], o: '', opt: '' }

The equivalent long option already behaves correctly, which is the tell that this is a short-option-only bug:

    parser(['--opt='])  // => { _: [], opt: '' }   (correct)

## Root cause

In the short-option branch of `lib/yargs-parser.ts`, `letters = arg.slice(1, -1).split('')` strips the trailing `=`. For `-o=` that leaves `letters = ['o']`, so the `letters[j + 1] === '='` guard never matches. Parsing falls through to `setArg('o', defaultValue('o'))` (boolean `true`) and then `key = arg.slice(-1)[0]` picks the trailing `=` up as a spurious key.

## Fix

`next` is already computed as `arg.slice(j + 2)` and equals exactly `'='` only on the final letter when the arg ends in a trailing `=`. Adding that as an alternative trigger routes `-o=` through the existing `=`-handling branch, where `value = arg.slice(j + 3)` correctly resolves to `''`:

    if ((letters[j + 1] && letters[j + 1] === '=') || next === '=') {

Non-empty values (`-o=x`) are unaffected -- they still match the original `letters[j + 1] === '='` condition.

## Tests

Added a case next to the existing empty-value short-option test covering both `-o=` (empty) and the `-o=x` control. Full suite: 363 passing, lint clean.

---
This change was prepared with AI assistance and reviewed by the author.

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -318,7 +318,7 @@ export class YargsParser {
         for (let j = 0; j < letters.length; j++) {
           next = arg.slice(j + 2)
 
-          if (letters[j + 1] && letters[j + 1] === '=') {
+          if ((letters[j + 1] && letters[j + 1] === '=') || next === '=') {
             value = arg.slice(j + 3)
             key = letters[j]
 
--- a/test/yargs-parser.mjs
+++ b/test/yargs-parser.mjs
@@ -186,6 +186,13 @@ describe('yargs-parser', function () {
     parse.should.have.property('_').with.length(0)
   })
 
+  // Addresses: https://github.com/yargs/yargs-parser/issues/367
+  it('should set a short option to an empty string when a trailing equals sign supplies no value', function () {
+    parser(['-o='], { alias: { o: 'opt' } }).should.deep.equal({ _: [], o: '', opt: '' })
+    // control: a non-empty value after the equals sign is unaffected
+    parser(['-o=x'], { alias: { o: 'opt' } }).should.deep.equal({ _: [], o: 'x', opt: 'x' })
+  })
+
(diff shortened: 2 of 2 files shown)


lbesecker195 (approved):
Traced `-o=`: `letters` is `['o']`, so `letters[1]` is undefined and the old check missed it, while `next` is `'='`. The added condition routes it to the existing `arg.slice(j + 3)` path, yielding `''`. Test covers both the empty and non-empty cases. Fixes #367. LGTM.

PR #542, opened 2026-08-25
Respect required args before array defaults

### LINKED ISSUE #445: array w/ default breaks requiresArg

wantosure:
Fixes #445.

When an array option had both a default and a required argument count, the missing-value path filled in the default before `narg` validation ran. That made a default look like a user-provided value, so `-a` could pass even though yargs asked for an argument.

This keeps counting the values that actually came from argv separately from the array value we eventually assign, so defaults still apply but they no longer satisfy the required-argument check.

Tests run:
- `npx mocha test/yargs-parser.mjs --grep "array with nargs|nargs"`
- `npm run check`
- `npm test`


shadowspawn:

PR #543, opened 2026-08-25
Let narg zero flags beat false defaults

### LINKED ISSUE #418: Default value of `false` doesn't work for `narg: { flag: 0 }`
`yargs-parser` doesn't seem to work when a boolean flag is defaulted to `false` and then activated from the args string. For example:

```javascript
const YargsParser = require('yargs-parser');

YargsParser('-a hello -b world', {
    narg: {
        a: 1,
        b: 0,
    },
    default: {
        b: false,
    },
    alias: {
        blah: [ 'b', 'bl' ],
    },
});


// expected output
{
    _: ['world'],
    a: 'hello',
    blah: true,
    b: true,
    bl: true
}


// actual output
{
    _: ['world'],
    a: 'hello',
    blah: false,
    b: false,
    bl: false
}
```

If I explicitly marked it as `boolean: true`, it seems to work as expected. However, the problem is if the option/yargs-parser configuration had `boolean` replaced with `narg` to allow the flag value to be of multiple types (falling back to a boolean if no value given), then setting it solely as boolean will break the functionality of the other values it could be.

Admittedly, `parsedArgs.b` would be `undefined` if it weren't defaulted which is still falsey, but this still came as a surprise to me since `narg: { blah: 0 }` should be enough to mark it as a boolean. It's not the end of the world but would be nice to not have this gotcha present in the parser that breaks if the user happens to add a default value of `false`.

**Related**
* [Issue 256](https://github.com/yargs/yargs/issues/256).
* [Issue 216](https://github.com/yargs/yargs-parser/issues/216).
* [PR 191](https://github.com/yargs/yargs-parser/pull/191).

wantosure:
Fixes #418.

A zero-argument option still needs to record that the flag was present. When the same option also had a `false` default, the old path reused `defaultValue(key)`, so the explicit `-b` ended up as `false` and its aliases picked up the same default.

This keeps the existing `number` and `string` behavior for `narg: 0`, but lets boolean-like zero-arg flags override the default with `true` when the user actually passes them.

Tests run:
- `npx mocha test/yargs-parser.mjs --grep "narg 0|zero narg|nargs"`
- `npm run check`
- `npm test`

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -465,7 +465,11 @@ export class YargsParser {
         if (!isUndefined(argAfterEqualSign)) {
           error = Error(__('Argument unexpected for: %s', key))
         }
-        setArg(key, defaultValue(key))
+        if (!checkAllAliases(key, flags.strings) && !checkAllAliases(key, flags.numbers)) {
+          setArg(key, true)
+        } else {
+          setArg(key, defaultValue(key))
+        }
         return i
       }
 
--- a/test/yargs-parser.mjs
+++ b/test/yargs-parser.mjs
@@ -1941,6 +1941,29 @@ describe('yargs-parser', function () {
       argv.error.message.should.equal('Argument unexpected for: bar')
     })
 
+    it('should allow narg 0 to override a false default', function () {
+      const argv = parser('-a hello -b world', {
+        narg: {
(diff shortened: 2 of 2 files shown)


shadowspawn:

PR #544, opened 2026-08-25
Keep quotes from config values intact

### LINKED ISSUE #385: String values in config files that contain embedded quotes are not handled correctly
If a config object specified via the `.config` option to yargs contains a value that starts and ends with quotes, yargs strips those quotes, due to this code here:

https://github.com/yargs/yargs-parser/blob/50b54ae66aed4240021ba8fed9e98fec3f8d07ef/lib/yargs-parser.ts#L608

That seems like confusing, and undesirable behavior, and makes it difficult to pass values containing embedded quotes.

For example:

```
console.log(require('yargs').config({a: JSON.stringify('hello')}).parse());
```

This prints:

```
{ _: [], a: 'hello', '$0': 'yargs-bug.js' }
```

But I would expect it to instead print:

```
{ _: [], a: '"hello"', '$0': 'yargs-bug.js' }
```

I would suggest that when parsing already-split arguments, e.g. from process.argv, or when processing a config file, quotes should *never* be stripped, and no additional splitting should ever occur.  The only time stripping of quotes should happen is when parsing an unsplit string.  In that case, the string should be parsed into an array of strings using shell-style parsing, and then parsing should proceed using the resultant array, and do no further splitting or quote handling.

wantosure:
Fixes #385.

Config values are already parsed before they reach `setConfigObject`, but that path was still using the default `setArg` quote stripping. A value like `"hello"` from JSON therefore became `hello`, which is surprising because no shell-style string tokenization is involved there.

This passes config values through without quote stripping, while leaving argv string parsing unchanged.

Tests run:
- `npm test -- --grep "config"`
- `npm run check`
- `npm test`

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -709,7 +709,7 @@ export class YargsParser {
           // setting arguments via CLI takes precedence over
           // values within the config file.
           if (!hasKey(argv, fullKey.split('.')) || (checkAllAliases(fullKey, flags.arrays) && configuration['combine-arrays'])) {
-            setArg(fullKey, value)
+            setArg(fullKey, value, false)
           }
         }
       })
--- a/test/fixtures/config.json
+++ b/test/fixtures/config.json
@@ -2,6 +2,7 @@
     "herp": "derp",
     "z": 55,
     "foo": "baz",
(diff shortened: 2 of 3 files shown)


shadowspawn:

PR #545, opened 2026-08-25
Treat empty array equals as a value

### LINKED ISSUE #368: Using an equal sign (or not) with an array option changes the value
When using the following code:

```js
const yargs = require('yargs')
console.log(yargs.option('opt', { array: true }).parse().opt)
```

Using a `=` or not changes the parsed value, which is quite confusing:

```bash
$ node yargs.js --opt=""
[]

$ node yargs.js --opt ""
[ '' ]
```

Using yargs `16.2.0` on Node `15.6.0`.

wantosure:
Fixes #368.

Array parsing treated `--opt=` differently from `--opt ""` because the empty string after `=` was selected with `||`. That made it fall through as if there was no value at all, so the array became `[]`.

This keeps an explicit empty string from the equals form and stores it the same way as the space-separated form.

Tests run:
- `npm test -- --grep "array|empty equals"`
- `npm run check`
- `npm test`

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -503,7 +503,7 @@ export class YargsParser {
     // e.g., --foo apple banana cat becomes ["apple", "banana", "cat"]
     function eatArray (i: number, key: string, args: string[], argAfterEqualSign?: string): number {
       let argsToSet = []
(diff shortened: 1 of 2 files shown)


shadowspawn:

PR #546, opened 2026-08-25
Strip dashed keys inside dot notation

### LINKED ISSUE #223: dot-notation doesn't respect strip-dashed
Hi!

With `"strip-dashed": true`

```
$ cli --dashed-will-be-stripped --here.it-will-not
```

```js
{
  dashedWillBeStripped: true,
  here: { 'it-will-not': true, itWillNot: true }
}
```

wantosure:
Fixes #223.

`strip-dashed` only looked at top-level argv keys. With dot notation, the dashed form inside a nested object stayed around next to the camel-cased key, so `--here.it-will-not` produced both `it-will-not` and `itWillNot`.

This makes the dashed-key cleanup walk nested plain objects as well, while still preserving the special `--` key.

Tests run:
- `npm test -- --grep "stripping|strip-dashed"`
- `npm run check`
- `npm test`

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -429,9 +429,7 @@ export class YargsParser {
     })
 
     if (configuration['camel-case-expansion'] && configuration['strip-dashed']) {
-      Object.keys(argv).filter(key => key !== '--' && key.includes('-')).forEach(key => {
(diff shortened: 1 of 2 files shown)


shadowspawn:

PR #547, opened 2026-08-29
fix: support object values in coerce handlers

### LINKED ISSUE #352: maybeCoerceNumber bombs if coerce produced an object without toString
I ran into this trying to use `pg-connection-string.parse` as a coerce function today, it returns an object with a null prototype which can't be coerced to a string. 

```typescript
import yargs from 'yargs';
import {parse as pgcs_parse} from 'pg-connection-string';


const parser = yargs('--db=postgresql://db/something'.split(/\s+/))
  .strict()
  .option('db', {
    coerce: pgcs_parse,
  })
  .fail((_, err) => {
    console.error(err.stack);
    process.exit(1);
  })
  ;

const argv = parser.parse();

console.log(argv);
```

```
:; ts-node test.ts
YError: Cannot convert object to primitive value
    at Object.parseArgs [as _parseArgs] (.../node_modules/yargs/build/index.cjs:2762:27)
    at Object.parse (.../node_modules/yargs/build/index.cjs:2260:31)
    at Object.<anonymous> (.../test.ts:16:21)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Module.m._compile (/usr/local/lib/node_modules/ts-node/src/index.ts:858:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Object.require.extensions.<computed> [as .ts] (/usr/local/lib/node_modules/ts-node/src/index.ts:861:12)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
```

The original stack trace is lost when it gets converted to YError, but the problem occurs in `looksLikeNumber` when the regexp is applied.  It seems to me that `maybeCoerceNumber` could check if value is a string before doing anything more.

amtbsl:
Fixes #352.

Avoid treating object-valued options as arrays during coercion. Coerce handlers now receive object values directly, with targeted regression coverage.

Validation: TypeScript compile and targeted object-coercion regression test.

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -637,7 +637,13 @@ export class YargsParser {
       return value
     }
 
-    function maybeCoerceNumber (key: string, value: string | number | null | undefined) {
+    function maybeCoerceNumber (key: string, value: unknown) {
+      // Coercion callbacks may intentionally return objects (for example,
+      // parsed connection strings). Only strings and numbers can be inspected
+      // for numeric syntax; passing other values to the regular expressions
+      // in `looksLikeNumber` can throw for objects without a prototype.
+      if (typeof value !== 'string' && typeof value !== 'number') return value
+
       if (!configuration['parse-positional-numbers'] && key === '_') return value
       if (!checkAllAliases(key, flags.strings) && !checkAllAliases(key, flags.bools) && !Array.isArray(value)) {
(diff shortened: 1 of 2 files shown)


shadowspawn:
To be clear: is this Pull Request entirely AI/LLM generated?

PR #548, opened 2026-08-30
fix: do not treat option value 1 as a count increment

### LINKED ISSUE #506: Multiple arguments are incremented if they are equal to 1.

official-burak:
Fixes #506.

`parser("-x 3 -x 1")` currently returns `{ x: 4 }` because `setKey` treats any value of `1` as a count step (`value === increment()`, and `increment()` returns 1). Count flags now use a unique sentinel instead of the number 1.

`-x 3 -x 1` is now `[3, 1]`. With `duplicate-arguments-array: false` the last value stays `1`. Existing count tests still pass.

### DIFF
--- a/lib/yargs-parser.ts
+++ b/lib/yargs-parser.ts
@@ -626,7 +626,8 @@ export class YargsParser {
 
       // increment a count given as arg (either no value or value parsed as boolean)
       if (checkAllAliases(key, flags.counts) && (isUndefined(value) || typeof value === 'boolean')) {
-        value = increment()
+        // COUNT_INCREMENT is a unique sentinel, not the number 1.
+        value = COUNT_INCREMENT as unknown as typeof value
       }
 
       // Set normalized value when key is in 'normalize' and in 'arrays'
@@ -850,7 +851,7 @@ export class YargsParser {
         }
       }
 
-      if (value === increment()) {
+      if (value === COUNT_INCREMENT) {
         o[key] = increment(o[key])
       } else if (Array.isArray(o[key])) {
         if (duplicate && isTypeArray && isValueArray) {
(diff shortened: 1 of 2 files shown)


shadowspawn:
This looks like a drive-by AI contribution. The account created 29 other Pull Requests in 12 days.

This may get looked at and used when the issue is prioritised.

The human maintainer does not have time to review all AI heavy PRs that are opened.

official-burak:
> This looks like a drive-by AI contribution. The account created 29 other Pull Requests in 12 days.
> 
> 
> 
> This may get looked at and used when the issue is prioritised.
> 
> 
> 
> The human maintainer does not have time to review all AI heavy PRs that are opened.

Thanks for your feedback. I focus on AI-Native development, so I heavily utilize AI in my coding processes. However, I stand behind the code I submit and believe it adds value to the project.

I completely understand the time constraints human maintainers face. I respect your decision to close this PR; feel free to revisit it whenever the issue is prioritized.

